PR #10, opened 2019-11-30
Initial support for setup and interactive commands, using powershell in Windows

coreyeng:
Hi,

I got O2 working in Windows. There were some problems, but I think I've at least the basics worked out. With the updates here, and the few additional steps added to the README, I can get the `setup` and `interactive` commands to work.

I think there's still some improvements to make here but this should at get some Windows-side developments going and, given the volatile nature of this project being very early in development, I don't really want to sit on these changes while I continue to learn Rust and risk having merging nightmares later on.

I don't have WSL installed yet, but I'll be doing that pretty soon here and I can start running O2 on both systems. I don't think anything I changed should have affected Linux.

I can give some more details if needed, but most changes boil down to syntax differences between Linux and Windows, and then how Rust handles system commands, which IMO, is really user-unfriendly since no other languages I'm aware of calls commands in that way. Not even C++ or Java.

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 *.egg-info
 __pycache__
 /rust/origen/target
+pip-wheel-metadata/
+_origen.pyd
 
 # Editor cruft
 *.swp
--- a/README.md
+++ b/README.md
@@ -1,5 +1,14 @@
 ## Development Environment Setup
 
+0) Windows-Specific Setup
+
+If using Windows, ensure that the `HOME` environment variable is set. For example, using Powershell:
+
+~~~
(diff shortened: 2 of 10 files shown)


coreyeng:
Oh, forgot to mention, Python's `readline` isn't supported in Windows. I found a pure-python implementation, [pyreadline](https://pypi.org/project/pyreadline/), but it doesn't support the history stuff. I left it in for now if we want to play with it, but I wanted to get some basic Windows supported added. For now, I just added warnings that the history file isn't supported. I don't think it'd be too complicated to homebrew our own, but if we start using other things in the `readline` library, we may have to differentiate between Windows and Linux.

PR #14, opened 2019-12-05
Logger and windows history file

coreyeng:
Hi,

Here's an attempt at getting a logger going. I'd like to add some customization yet but this should suffice for now.

The logger is available both on the Rust and Python side. I changed the `version` command to use the logger, so there's an example from the Rust side there:

image

On the Python side, the logger is available at `origen.logger` and has methods `debug, deprecated, error, info, log, success, warning` for logging and `output_file` to retrieve the (currently hard-coded) filename.

I added a block syntax to the logger which will align the text but you can pass a newline as before format yourself. For example, in Python:

image

One change from the Ruby implementation logger is that I made `info` stand out a bit more by printing it in cyan. The version command will now print as:

image

To get plain text out of the logger, `log` is available:

image

Like I said, there's still some customization options to be added, but I think this should be enough to get started. We can either discuss customizations here or in the project board.

I also got the history file on Windows to work. On the plus side: it works, but on the minus side, its kinda bulky and I've not had any success in extracting it out or cutting it down. I'm not quite sure why and I can look at it a bit more, especially if `boot.py` grows more but its tricky to debug since my attempts would appear to work at first, but then the interpreter just crashes randomly, without any error or stack trace. The setup was taken directly from the `pyreadline` [documentation](https://pythonhosted.org/pyreadline/introduction.html) so I'm tempted to just leave it as it is unless it gets too in-the-way - another point which can discuss here or in the project board.

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ __pycache__
 /rust/origen/target
 pip-wheel-metadata/
 _origen.pyd
+log/
 
 # Editor cruft
 *.swp
--- a/python/origen/__init__.py
+++ b/python/origen/__init__.py
@@ -7,6 +7,7 @@ config = _origen.config()
 status = _origen.status()
 root = Path(status["root"])
 version = status["origen_version"]
+logger = _origen.logger
 
 app = None
 dut = None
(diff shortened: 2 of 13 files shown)


coreyeng:
Forgot to double check the out file. Bugfix above has properly aligned text now:

image

image

Also changed 'INFO' to 'LOG' in `logger.log`

PR #17, opened 2019-12-08
Basic block loader

ginty:
Sorry that this is a quite a wide ranging PR, a lot of these early features are closely coupled.

It's of course not finished yet, but I wanted to get it out ASAP since it adds core infrastructure upon which things like the pins model and APIs can be built.

It adds the following:

- Working sub_blocks.py file loading
- Working registers.py file loading
- Defines how all the DUT data should be stored in Rust in a way that will support target switching
- Defines how Rust errors should be handled

## DUT data storage

For background on how the DUT data is stored and why, read this - https://github.com/Origen-SDK/o2/wiki/DUT-Data-Storage

The top-level DUT struct just instantiates a single Model stuct and then models can instantiate other models to an arbitrary depth.
Pins and other attributes should be added to Model - https://github.com/Origen-SDK/o2/blob/block_loader/rust/origen/src/core/model/mod.rs#L27

There is a method to get a handle on an embedded model based on a path (like 'core0.adc0') which will be passed in from the Python side, here is an example - https://github.com/Origen-SDK/o2/blob/block_loader/rust/origen/pyapi/src/dut.rs#L52

## Block files loading

Each type of block file (registers.py, pins.py, sub_blocks.py, etc.) should have a Loader class which is responsible for implementing a user-friendly API to define the given attribute and then translate that into calls to the Rust DUT object.

The loading of each file type is lazy and kicked off from here - https://github.com/Origen-SDK/o2/blob/block_loader/python/origen/controller.py#L28

A proxy object shadows the database on the python side to provide user-friendly access to the underlying raw data store. This is very much in its infancy, but you can see the implementations for these for the regs and sub-blocks here:

- https://github.com/Origen-SDK/o2/blob/block_loader/python/origen/registers.py
- https://github.com/Origen-SDK/o2/blob/block_loader/python/origen/sub_blocks.py

I'm not sure how far this shadowing of the raw data will go, currently it is just for collections (<my_block>.sub_blocks and <my_block>.regs), but possibly this will be extended futher to creating such shadow objects down to lower levels like registers.

These APIs work from the console:

~~~python
dut.sub_blocks["core2"].sub_blocks["adc1"]             # => <'adc1' controller>
dut.sub_blocks["core2"].sub_blocks["adc1"].regs        # => <regs proxy>
dut.sub_blocks["core2"].sub_blocks["adc1"].regs.len()  # => 1
# Number of regs owned by the DUT top-level
dut.regs.len()   # => 3
~~~

## Rust error handling

Any errors generated by Rust should be piped all the way back to the Python API and allowing Rust to panic should be avoided.
A panic error message will mean nothing to the user and if we do our jobs right the only way for Rust to error will be when the user input is at fault in some way.

By contrast, allowing the error to bubble up and back to Python will result in a Python stack trace being generated along with the error message. That gives the user much more feedback about where in their code they have gone wrong.

You can see examples of this by modifying the sub block or register files in example to create a reg/block with a duplicate name. Go into the console and use the above APIs to load the given resource and you will see a Python error with an error message that was originated in the Rust code.

Rust's error handling is actually pretty nice and you should read up on it if you haven't already. There are some nice shorthands for bubbling errors like this.

I added an `origen::error::Error` type (should probably alias that to `origen::Error`) and you should use that anytime you are manually creating an error, here is an example - https://github.com/Origen-SDK/o2/blob/block_loader/rust/origen/src/core/model/mod.rs#L145

We might need to add conversions so that Origen errors can be cast to other errors, like this - https://github.com/Origen-SDK/o2/blob/block_loader/rust/origen/src/error.rs#L49
However, annoyingly I haven't been able to get to work for the PyErr type that we need to return to Python, so you will see them manually being converted in the dut pyapi functions right now.

For functions that can generate an `origen::Error`, you should have them return type `origen::Result<type>`, which is a shorthand for `Result<type, origen::Error>`, here is an example - https://github.com/Origen-SDK/o2/blob/block_loader/rust/origen/src/core/dut.rs#L21

I believe that this approach to error handling is idiomatic Rust and it doesn't get us invovled with any error handling crates since there doesn't seem to be any de-facto standard for that yet.


coreyeng:
Looks good! I'm doing something similar for the pins, except I didn't call it `db` and I put them on the `toplevel` instead of `model`.

Are we using the wiki for the docs for O2 then? Instead of README files?

coreyeng:
It looks like the formatting was changed quite a bit as well. Did you run a linter over the project? I don't anything in the tomls that looks like a linter though.

ginty:
Hi @coreyeng,

> Are we using the wiki for the docs for O2 then? Instead of README files?

Not instead of user docs, but I wanted to start adding some developer information and it seems like a good place for that. The user docs should be on a new version of origen-sdk.org in future and I think we should keep developer-type info strictly separate to keep the user info as concise as possible.

> Did you run a linter over the project?

Yes, it's part of Rust, run `cargo fmt`

coreyeng:
@ginty Do the test in the example app pass? I'm seeing them all fail. I'll look into it a bit but just want to see if you were still using them or if maybe something didn't get checked in.

ginty:
@coreyeng, no sorry, I just added them to give me something to develop against but then stopped using them and forgot about them - so not expected to pass in their current state.
Just ignore them for now and I'll fix soon, thanks.

ginty:
Hi @coreyeng,

OK I checked in working tests to master.

One thing I noticed was that giving a module access to dut like this doesn't work:

~~~python
from origen import dut

dut.blah
~~~

That fixes the dut reference after the first time it is imported and future instantiations of a class from that module do not pick up the new dut if it has been changed.

This does work though:

~~~python
import origen

origen.dut.blah
~~~

Need to think through what changes should be made to prevent the first case from being possible.

coreyeng:
Thanks! I had just commented them out and forced the DUT for my pin tests. Hopefully I'll have something for pins pretty soon.

ginty:
@coreyeng 

> I'm doing something similar for the pins, except I didn't call it db and I put them on the toplevel instead of model.

I think one benefit of putting them in model (although only using the top-level model for now) is that it will be an architecture that could in future support a multi-chip module. In that case the top-level model would be a module with the top-level pins, and then the first layer of sub-blocks would be chips with their own pin definitions.

We'd need to add something else to model the connections between the pins, but at least having the ability to create pins at different levels is good enough for now.

Issue #20, opened 2019-12-10
Crash on Windows when outside of workspace


pderouen:
https://github.com/Origen-SDK/o2/blob/64e584707e8736df7107db8d0ffcfcfc639e8190/rust/origen/src/core/utility/logger.rs#L128-L131

The HOME environment variable does not exist by default in Windows.

coreyeng:
See `step 0` of the README, which addresses this. O2 isn't setting `HOME` if it doesn't exist, and I'm really not sure if it even should.

pderouen:
Could we not instead define a home equivalent if running under Windows? Most tools you use don't require environment variables to be setup. Windows can already point you to the current user's document folder (for example).

If possible, this time around, it would be great if functionality on Windows was seamless and "native" instead of Linux-ified. :)

ginty:
Also consider that if it can't work out a HOME then it should maybe log to the console only and emit a warning that it is not logging to a file because it can't work out what the user home dir is

pderouen:
Good point

PR #26, opened 2019-12-13
Log dir - Windows update

pderouen:
This is to fix https://github.com/Origen-SDK/o2/issues/20

It took quite a while to get the code to compile... I'm really looking forward to getting the hang of variable ownership.

### DIFF
--- a/rust/origen/cli/src/python.rs
+++ b/rust/origen/cli/src/python.rs
@@ -1,6 +1,7 @@
 // Responsible for managing Python execution
 
 use origen::core::os;
+use origen::STATUS;
 use semver::Version;
 
 const PYTHONS: &[&str] = &[
@@ -34,7 +35,7 @@ impl Default for Config {
                     available = true;
                     let poetry_cmd = format!(
                         "{}/.poetry/bin/poetry",
-                        std::env::var("HOME").expect("$HOME is not defined")
+                        format!("{}", STATUS.home.to_string_lossy())
                     );
                     if version >= Version::parse(MIN_PYTHON_VERSION).unwrap() {
                         return Config {
--- a/rust/origen/cli/src/python.rs
+++ b/rust/origen/cli/src/python.rs
@@ -35,7 +35,7 @@ impl Default for Config {
                     available = true;
                     let poetry_cmd = format!(
                         "{}/.poetry/bin/poetry",
-                        format!("{}", STATUS.home.to_string_lossy())
+                        format!("{}", STATUS.home.display())
                     );
                     if version >= Version::parse(MIN_PYTHON_VERSION).unwrap() {
                         return Config {
--- a/rust/origen/src/core/status.rs
+++ b/rust/origen/src/core/status.rs
@@ -63,9 +63,9 @@ fn search_for_app_root() -> (bool, PathBuf) {
 
 fn get_home_dir() -> PathBuf {
     if cfg!(windows) {
-        PathBuf::from(env::var("USERPROFILE").expect("No environment variable for USERPROFILE"))
+        PathBuf::from(env::var("USERPROFILE").expect("Please set environment variable USERPROFILE to point to your home directory, then try again"))
     }
     else {
-        PathBuf::from(env::var("HOME").expect("No environment variable for HOME"))
+        PathBuf::from(env::var("HOME").expect("Please set environment variable HOME to point to your home directory, then try again"))
     }
 }
--- a/rust/origen/src/core/status.rs
+++ b/rust/origen/src/core/status.rs
@@ -17,6 +17,8 @@ pub struct Status {
     /// The Origen version in a Semver object
     pub origen_version: Version,
     pub start_time: time::Tm,
+    /// The full file system path to the user's home directory
+    pub home: PathBuf,
 }
 
 impl Default for Status {
@@ -31,6 +33,7 @@ impl Default for Status {
             root: r,
             origen_version: version,
             start_time: time::now(),
+            home: get_home_dir(),
         }
     }
 }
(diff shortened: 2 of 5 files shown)


pderouen:
@coreyeng, should I go ahead and remove the log dir creation code in <code>fn default()</code>?

coreyeng:
I think it's still need. I had two log files setup: one that would be in `home/.origen/log` for running globally (even though we don't have much to run globally yet) and one for the application, which is at `<app_root>/log`. Creating the log directories if they don't exists are still required. I don't think there's a "create file recursively" function.

coreyeng:
I think too we can remove `step 0` of the README now that setting a `HOME` isn't required.

Would it also be good to save off the home directory as `origen::core::HOME` or something like that? To ensure that we have an OS-agnostic way to grabbing the home directory? Eventually we'll probably want a user class to wrap all that, but for now it may be good. I think we had `Origen.home` in O1 for this, with the `.origen` attached. That is, `<home dir>.origen`.

pderouen:
> I think too we can remove step 0 of the README now that setting a HOME isn't required.

Will do

> Would it also be good to save off the home directory as origen::core::HOME or something like that? To ensure that we have an OS-agnostic way to grabbing the home directory? Eventually we'll probably want a user class to wrap all that, but for now it may be good. I think we had Origen.home in O1 for this, with the .origen attached. That is, <home dir>.origen.

I'm not sure what you mean by save off the home directory. For now I'm grabbing it from the windows equivalent to the home var.

I updated the get dir fn again to always create any missing dirs before returning the path. It works now from a workspace. Turns out <code>fs::create_dir_all</code> will recursively create all missing dirs.

pderouen:
For clarity on windows the log file will now land in <code>c:\\users\\user_name\\.origen\\log\\</code> by default when running outside of a workspace. I chose that because o1 stuck a .origen directory there. Rust stuck stuff there too. So, seemed like the place to be!

ginty:
Saving the home seems like a good idea, i.e. work it out once and then save it to here - https://github.com/Origen-SDK/o2/blob/master/rust/origen/src/core/status.rs
Then everywhere else in the app can just reference `origen::STATUS.home`

coreyeng:
@pderouen, yeah, an `origen::STATUS.home` or something to that effect is what I meant. It'd be a nice way for other features to grab the same directory as well.

Though, that'd be a bonus item so I'm okay with taking this PR as is.

pderouen:
I'm on board with making that update in this PR. It'll give me another chance to fight with (learn) the borrow checker.

ginty:
@pderouen, just to give you a hint - note that `origen::STATUS` is immutable, so the only way to set the value of this is from within the default function.

pderouen:
Yup, I've got that all working. But, I'm tracking down the source of a panic when I run <code>origen i</code>:

<code>
thread 'main' panicked at '$HOME is not defined: NotPresent', src\libcore\result.rs:1189:5
</code>

pderouen:
I'm having trouble locating that file

pderouen:
nevermind, backtrace set me straight

pderouen:
The source of the panic is here: https://github.com/Origen-SDK/o2/blob/c4662b5fb677b4e7d265e71e89f0a17fbd9687ab/rust/origen/cli/src/python.rs#L37

I can update it to pull home from status. But, I still get a run time error on <code>origen i</code>. I working from a user account that has a space in the name which is resulting in:

<code>
'C:\Users\Paul' is not recognized as an internal or external command,
operable program or batch file.
</code>

I'll check in the update to pull home from status. I'm not sure how to fix the path with spaces issue as I'm not sure what that poetry command is doing or where it's used.

pderouen:
I've figured out the cause of my <code>origen i</code> issue. I have a fix that works for me. But, I think it belongs in a separate thread. I'll raise an issue.

@ginty, @coreyeng, are you ok with is PR as is?

rust/origen/src/core/utility/logger.rs
```diff
@@ -153,14 +138,15 @@ impl Logger {
...
+            pb = (STATUS.root).to_path_buf().join("log");
         } else {
-            match env::var("HOME") {
-                Ok(var) => PathBuf::from(var).join(".origen").join("log"),
-                Err(_e) => panic!("No environment variable for HOME!"),
-            }
+            pb = (STATUS.home).to_path_buf().join(".origen").join("log");
         }
+        // create all missing directories to avoid panics
+        fs::create_dir_all(pb.as_path()).expect("Could not create the log directory");
```
ginty: Should we include the directory that it is trying to make in the error message?
Any panics like this from Rust will just show the user this message and nothing else, so they need to be as descriptive and helpful as possible.

rust/origen/src/core/status.rs
```diff
@@ -57,3 +60,12 @@ fn search_for_app_root() -> (bool, PathBuf) {
         (true, path)
     }
 }
+
+fn get_home_dir() -> PathBuf {
+    if cfg!(windows) {
+        PathBuf::from(env::var("USERPROFILE").expect("No environment variable for USERPROFILE"))
+    }
+    else {
+        PathBuf::from(env::var("HOME").expect("No environment variable for HOME"))
```
ginty: I think we should phrase these as "Please set environment variable HOME to point to your home directory, then try again"

rust/origen/cli/src/python.rs
```diff
@@ -34,7 +35,7 @@ impl Default for Config {
                     available = true;
                     let poetry_cmd = format!(
                         "{}/.poetry/bin/poetry",
-                        std::env::var("HOME").expect("$HOME is not defined")
+                        format!("{}", STATUS.home.to_string_lossy())
```
ginty: Is there a reason for `to_string_lossy()` here and not `display()` as used elsewhere?
Could be, I'm just curious as I haven't came across this method yet.

pderouen: I found it in other code. I just needed a method that would satisfy the borrow checker. @coreyeng must be the one who uses that method.

pderouen:
Updated as requested

PR #39, opened 2019-12-21
Regs WIP

ginty:
This is partial work on implementing the regs APIs, which mainly adds the ability to instantiate regs within memory map and address block structures.

It also adds friendly DUT traversal (`dut.core0.adc0.memory_maps.user.bank0.regs`) and a `tree()` method for looking at the dut structure in the console (`dut.tree()`, `dut.core0.tree()`).

I would like to merge it at this point as I am about to experiment with changing how the DUT object storage works from a hierarchical representation to some kind of ID-based system and I would like to de-couple that change from this stuff that I have in progress.

Here's the API example for defining regs in a block's registers.py file:

~~~python
# Registers added in the global scope of these files will be added to
# a memory map called 'default' and within that an address block called
# 'default'. Such regs can be accessed via my_block.regs and users who
# don't care about maps don't have to even think about them.
# Note that such regs can also be accessed via my_block.default.default.regs

# A simple reg definition with all bits writable, here at address 0 and a
# default size of 32-bits
SimpleReg("reg1", 0)
# Another simple reg with custom size
SimpleReg("reg2", 4, size=16)

# This is the reg definition
with Reg("reg3", 0x0024, size=16) as reg:
    # This is the COCO definition
    reg.bit(7, "coco", access="ro")
    reg.bit(6, "aien")
    reg.bit(5, "diff")
    reg.bit([4,0], "adch", reset=0x1F)

# Regs can be added within a defined memory map, and in this case no address
# block is given so that will mean they are placed in a default address block
# named 'default'.
with MemoryMap("user"):
    # Test that reg names can be reused when scoped within a different map
    SimpleReg("reg1", 0)

    with Reg("reg2", 0x0024, size=16) as reg:
        reg.bit([4,0], "adch", reset=0x1F)


# Finally regs can be added to a full declared scope like this:
with MemoryMap("test"):
    with AddressBlock("bank0"):
        # Test that reg names can be reused when scoped within a different map
        SimpleReg("reg1", 0)

        with Reg("reg2", 0x0024, size=16) as reg:
            reg.bit([4,0], "adch", reset=0x1F)
~~~

### DIFF
--- a/example/example/blocks/adc/registers.py
+++ b/example/example/blocks/adc/registers.py
@@ -1,1 +1,1 @@
-reg("reg1", 0)
+SimpleReg("reg1", 0)
--- a/example/example/blocks/dut/derivatives/falcon/registers.py
+++ b/example/example/blocks/dut/derivatives/falcon/registers.py
@@ -1,1 +1,1 @@
-reg("reg3", 0x20)
+SimpleReg("freg1", 0x20)
--- a/example/example/blocks/dut/registers.py
+++ b/example/example/blocks/dut/registers.py
@@ -21,9 +21,12 @@ with Reg("reg3", 0x0024, size=16) as reg:
 # Regs can be added within a defined memory map, and in this case no address
 # block is given so that will mean they are placed in a default address block
 # named 'default'.
-with MemoryMap("user") as mem:
-    pass
+with MemoryMap("user"):
+    # Test that reg names can be reused when scoped within a different map
+    SimpleReg("reg1", 0)
 
+    with Reg("reg2", 0x0024, size=16) as reg:
+        reg.bit([4,0], "adch", reset=0x1F)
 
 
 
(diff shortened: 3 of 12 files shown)


info-rchitect:
Perfect timing as I am making progress on the IP-XACT importer, approved.

coreyeng:
Looks good to me, but can you expand a bit on:

>  I am about to experiment with changing how the DUT object storage works from a hierarchical representation to some kind of ID-based system

Do you mean storing the objects in basically a lookup table with IDs instead of a tree-like structure that we have now?

ginty:
Hi @coreyeng, yes exactly that.
Storing it in an OO fashion was getting increasingly convoluted as I wanted to access more deeply nested objects like regs and bits. I'm close to getting a PR for the new system for you to look at.

info-rchitect:
What will be the performance impact for this ID system in general?  At first glance it sounds like it would be improved versus hierarchical.

ginty:
> What will be the performance impact for this ID system in general? At first glance it sounds like it would be improved versus hierarchical.

Yeah it will be better, though I'm not sure if it would really be significant or not.

PR #49, opened 2020-01-02
Pins wip

coreyeng:
Some work on reorganizing and cleaning up the pins implementation from #21. I've got both mutable and immutable accessors and did away with some unnecessary copying. Per #40, I also rebranded `id` to `name`, so the API changed slightly, but nothing too major. Basically just a find & replace from `id` to `name`.

This appears as though there's a lot of changes, but the TL;DR is:
* `id` is now `name`
* Pins implementation split out to multiple files.

### DIFF
--- a/example/tests/pin_api_test.py
+++ b/example/tests/pin_api_test.py
@@ -818,6 +818,13 @@ def test_collecting_with_mixed_inputs():
   c = origen.dut.pins.collect("/port.1/", "p1", r)
   assert c.pin_names == ["porta1", "portb1", "p1", "porta0", "portb0"]
 
+# def test_pins_have_empty_metadata():
+#   assert origen.dut.pin("porta1").metadata == {}
+#   assert origen.dut.physical_pin("porta1").metadata == {}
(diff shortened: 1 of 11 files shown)


ginty:
Looks good to me

coreyeng:
@ginty Can I revert the cargo fmt that was push ~30 minutes ago? That really screwed with this PR since everything is moved around. I'm basically going to have to redo the PR since Git will see everything as a conflict.

coreyeng:
On second thought, I can just ignore master on conflicts... I'll redo the format after its merged.

ginty:
Crap, sorry. I did a cargo fmt by accident on #51 and then did the format on master to make that diff easier, but didn't think about the effect on this guy.

We can definitely revert it if it helps you, let me know if you need me to do anything.

coreyeng:
I just checked out my own and ran the formatter myself. I think the net result should be the same.

If I merge this, it should be an easy merge for you for other branches, right? I think it'll recognize moving stuff around alright since I'm up to date with master now.

ginty:
@coreyeng, sure, no problem.

Issue #50, opened 2020-01-02
Use of Python 3.8 syntax


ginty:
Master currently uses syntax only available in Python 3.8 and above (`:=`) - https://github.com/Origen-SDK/o2/blob/master/python/origen/controller.py#L13

However, O2 is otherwise configured as Python 3.6 to be the min version.

I originally chose 3.6 as the min version since that is the version installed by default in Ubuntu's latest LTS version (18.04), so that seemed to strike the right balance between allowing us to use modern Python features yet making environment setup straightforward.

So either we need to remove this code, or else discuss bumping the min version to 3.8.

ginty:
Agreed to stick with Python 3.6 as our min version

Issue #47, opened 2019-12-31
regs.keys() is not deterministic


info-rchitect:
Hi,

Here is my unit test:

~~~python
origen.app.instantiate_dut("dut.falcon")
    remote_file = f"{origen.root}/vendor/ip-xact/spirit1-4_ip-xact.xml"
    assert len(origen.dut.memory_maps) == 3
    origen.app.translate(remote_file)
    assert len(origen.dut.memory_maps) == 4    
    assert origen.dut.memory_map("RegisterMap").regs.len() == 2    
    assert origen.dut.memory_map("RegisterMap").regs.keys() == ['dut_top_level_reg_number_two', 'dut_top_level_reg']
    breakpoint()
~~~

The last unit test will toggle between passing and failing when poetry is restarted:

~~~shell
info-rchitect@D10-brian:/mnt/c/o2_dev/example$ poetry run pytest tests/translator_test.py
========================================== test session starts ==========================================
platform linux -- Python 3.8.1, pytest-3.10.1, py-1.8.1, pluggy-0.13.1
rootdir: /mnt/c/o2_dev/example, inifile: setup.cfg
collected 1 item

tests/translator_test.py F                                                                        [100%]

=============================================== FAILURES ================================================
________________________________________ test_translator_ip_xact ________________________________________

    def test_translator_ip_xact():
        origen.app.instantiate_dut("dut.falcon")
        remote_file = f"{origen.root}/vendor/ip-xact/spirit1-4_ip-xact.xml"
        assert len(origen.dut.memory_maps) == 3
        origen.app.translate(remote_file)
        assert len(origen.dut.memory_maps) == 4        
        assert origen.dut.memory_map("RegisterMap").regs.len() == 2
>       assert origen.dut.memory_map("RegisterMap").regs.keys() == ['dut_top_level_reg_number_two', 'dut_top_level_reg']
E       AssertionError: assert ['dut_top_lev...g_number_two'] == ['dut_top_leve...op_level_reg']
E         At index 0 diff: 'dut_top_level_reg' != 'dut_top_level_reg_number_two'
E         Use -v to get the full diff

tests/translator_test.py:14: AssertionError
======================================= 1 failed in 0.29 seconds ========================================
info-rchitect@D10-brian:/mnt/c/o2_dev/example$ poetry run pytest tests/translator_test.py
========================================== test session starts ==========================================
platform linux -- Python 3.8.1, pytest-3.10.1, py-1.8.1, pluggy-0.13.1
[...]

PR #51, opened 2020-01-02
Fix for #47

ginty:
This changes all instances of HashMap used in the model and register-related structs to be IndexMaps instead.

This means that keys and iterators based on these will return in definition order, which is the same as O1 and is probably the expected behavior by most users.
The application can sort if they want it in some other order, e.g. alphabetically.

@coreyeng, please push to this branch how you want to handle it for pins - I wasn't sure whether just to implement it across the board as I've done for regs, or if you preferred to do it only for the subset of user-facing cases.

Note also that this contains a small edit per #50 to get it working with my Python 3.6 env.

Also there are two pin tests failing (related to the regex feature), I'm not sure if these were already failing or if related to this edit or if related to me otherwise running py 3.6 instead of 3.8.

### DIFF
--- a/python/origen/controller.py
+++ b/python/origen/controller.py
@@ -10,7 +10,8 @@ class Proxies:
         self.proxies = {}
     
     def __getitem__(self, name):
-        if (p := self.proxies.get(name)):
+        p = self.proxies.get(name)
+        if (p):
             return p
         else:
             origen.logger.error(f"No proxy for '{name}' has been set!")
--- a/rust/origen/Cargo.lock
+++ b/rust/origen/Cargo.lock
@@ -871,6 +871,7 @@ version = "2.0.0-pre0"
 dependencies = [
  "built 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "config 0.9.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "indexmap 1.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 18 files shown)


ginty:
Hi @coreyeng,

Here are the fails:

~~~
___________________________________________________________________________________ test_collecting_with_single_regex ___________________________________________________________________________________
    def test_collecting_with_single_regex():
      import re
      r = re.compile("port.0")
>     c = origen.dut.pins.collect(r)
E     TypeError

tests/pin_api_test.py:808: TypeError
___________________________________________________________________________________ test_collecting_with_mixed_inputs ___________________________________________________________________________________
    def test_collecting_with_mixed_inputs():
      import re
      r = re.compile("port.0")
>     c = origen.dut.pins.collect("/port.1/", "p1", r)
E     TypeError

tests/pin_api_test.py:818: TypeError
~~~

I also see a compiler warnings now in master:

~~~
warning: trait objects without an explicit `dyn` are deprecated
   --> src/core/model/pins/pin.rs:112:35
    |
112 |     pub meta: HashMap<String, Box<Any + std::marker::Send>>,
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^ help: use `dyn`: `dyn Any + std::marker::Send`
    |
    = note: `#[warn(bare_trait_objects)]` on by default
~~~

coreyeng:
Everything has still been passing for me with 3.8. I'll install 3.6 and see if I can see the failures.

The entire scheme I was trying that generated that warning is gone in #53. It should build cleanly after that.

PR #52, opened 2020-01-02
Added programmatic add_sub_block API

ginty:
@info-rchitect this gives the missing API to add sub_blocks programmatically for your importer.

I also changed the method in sub_blocks.py files to be `SubBlock` to align with latest regs API and to give the user the impression that they are instantiating an object

@coreyeng, we need to build out the pins Loader at some point (just a gentle reminder, no rush but @info-rchitect will be looking for it eventually). We should follow a similar API as established by regs and sub-blocks, i.e. `Pin("my_pin")`

One thing I just noticed is that in my test I had to do:

~~~python
assert list(origen.dut.sub_blocks.keys()) == ['core0', 'core1', 'core2', 'core3']
~~~

whereas in the similar regs test @info-rchitect used:

~~~python
assert origen.dut.regs.keys() == ['reg0', 'reg1', 'reg2']
~~~

The difference is that the sub_blocks `keys()` method is implemented in Python and the others are implemented in Rust.
It seems that `keys()` returns an iterator view in Python 3 - https://stackoverflow.com/questions/31837161/python-difference-between-listdict-and-dict-keys/31837256

I'm not sure if it will be possible to return a similar iterator from the Rust-implemented dict-likes or not, though we could return a list from the `sub_blocks.keys()` for consistency if we wanted to.

 Is this something you have noticed @coreyeng?

### DIFF
--- a/example/example/blocks/core/sub_blocks.py
+++ b/example/example/blocks/core/sub_blocks.py
@@ -1,2 +1,2 @@
-sub_block("adc0", block_path="adc.16_bit")
-sub_block("adc1", block_path="adc.8_bit")
+SubBlock("adc0", block_path="adc.16_bit")
+SubBlock("adc1", block_path="adc.8_bit")
--- a/example/example/blocks/dut/sub_blocks.py
+++ b/example/example/blocks/dut/sub_blocks.py
@@ -1,4 +1,4 @@
-sub_block("core0", block_path="core")
-sub_block("core1", block_path="core")
-sub_block("core2", block_path="core")
-sub_block("core3", block_path="core")
+SubBlock("core0", block_path="core")
+SubBlock("core1", block_path="core")
+SubBlock("core2", block_path="core")
+SubBlock("core3", block_path="core")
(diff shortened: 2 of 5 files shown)


coreyeng:
> we need to build out the pins Loader at some point

No problem. It hasn't been a priority for me but it should be pretty straightforward since the backend stuff is there any working. I can get that added soon.

> I'm not sure if it will be possible to return a similar iterator from the Rust-implemented dict-likes or not, though we could return a list from the sub_blocks.keys() for consistency if we wanted to.
> 
> Is this something you have noticed @coreyeng?

I've implemented some iterators [in the containers](https://github.com/Origen-SDK/o2/blob/master/rust/origen/pyapi/src/pins/physical_pin_container.rs#L118-L166). I'm actually returning a `vec` from `keys` though so I'd imagine its getting converted to a `list` directly. I'd imagine I'd see the same. It returns some `dict_keys` thing that looked like a `list` to me, so I went with that.

info-rchitect:
@ginty is this ready to merge?

Issue #58, opened 2020-01-03
Export file structure of to_python method


info-rchitect:
Hi,

Can we clearly define the file structure created for a future `to_python` method?  This info will also be helpful when also making the associated `to_rust` method.  In O1 we only output pins, packages, sub_blocks, and registers.  We had no concept of memory maps or address blocks IIRC.

thx

ginty:
Do we have an example IP-XACT we can refer to?
If you or @chrisnappi can provide something then it would be easier to define what should be produced from it.

info-rchitect:
> Do we have an example IP-XACT we can refer to?
> If you or @chrisnappi can provide something then it would be easier to define what should be produced from it.

https://github.com/Origen-SDK/o2/blob/translator3/example/vendor/ip-xact/spirit1-4_ip-xact.xml

ginty:
I was thinking of a more elaborate/real life example, but basically a component in IP-XACT is an Origen block.
So that example would just generate a single block (a sub-directory in `<app>/blocks`) containing a register file like https://github.com/Origen-SDK/o2/blob/master/example/example/blocks/dut/registers.py

info-rchitect:
> I was thinking of a more elaborate/real life example, but basically a component in IP-XACT is an Origen block.
> So that example would just generate a single block (a sub-directory in <app>/blocks) containing a register file like https://github.com/Origen-SDK/o2/blob/master/example/example/blocks/dut/registers.py

So it wuld be `example/vendor/blocks/dut/registers.py`?

info-rchitect:
@ginty Also, do we need to have a sub-dir specifying `python` or would we potentially mix Python and Rust exported files in the same directory?

info-rchitect:
@ginty Also, what will the API to connect sub-blocks to memory maps and address blocks look like?  I believe we don't currently have a way to create a memory map and address block within a sub-block.

ginty:
> So it would be example/vendor/blocks/dut/registers.py?

Yeah, pretty much.

I think we might also add special handling for blocks which are DUTs (top-levels) vs. those which can be instantiated within a DUT, maybe the user will need to supply `dut=true` or similar at import-time.
We should follow what is done in `example/blocks` which is that all top-levels are below `blocks/dut`.
[...]

PR #59, opened 2020-01-03
To python

info-rchitect:
Using [this issue](https://github.com/Origen-SDK/o2/issues/58) as the basis for this PR.  I believe work is ongoing regarding the missing API to properly iterate through memory mpas, address blocks, registers.  Once that is firmed up I will add the necessary calls here.


info-rchitect:
@ginty going to re-open so the differences are more clear.

PR #64, opened 2020-01-06
added in custom error types

info-rchitect:
I think having some common error types will improve our error messaging over time.  We didn't really do this in O1 and I think it is highly encouraged in general.  Let me know if you disagree.

### DIFF
--- a/example/tests/errors_test.py
+++ b/example/tests/errors_test.py
@@ -3,13 +3,11 @@ import pytest
 from origen.errors import *
 
 def boot_falcon():
-    if origen.app is None:
-        return origen.app.instantiate_dut("dut.falcon")
-    else:
-        return origen.app
+    return origen.app.instantiate_dut("dut.falcon") if origen.dut is None else origen.dut
 
 def test_duplicate_detected_error():
-    app = boot_falcon()
-    with pytest.raises(DuplicateDetectedError) as err_info:
-        raise DuplicateDetectedError(app)
-    assert str(err_info.value) == "Cannot create instance of class 'example.application.Application' named 'example', it already exists!"
+    boot_falcon()
+    assert list(origen.dut.sub_blocks.keys()) == ['core0', 'core1', 'core2', 'core3']
+    with pytest.raises(DuplicateInstanceError) as err_info:
+        origen.dut.add_sub_block('core0')
+    assert str(err_info.value) == "Cannot create instance of 'sub_block' named 'core0', it already exists!"
--- a/example/tests/errors_test.py
+++ b/example/tests/errors_test.py
@@ -1,0 +1,15 @@
+import origen
+import pytest
+from origen.errors import *
+
+def boot_falcon():
+    if origen.app is None:
+        return origen.app.instantiate_dut("dut.falcon")
+    else:
+        return origen.app
+
+def test_duplicate_detected_error():
+    app = boot_falcon()
+    with pytest.raises(DuplicateDetectedError) as err_info:
+        raise DuplicateDetectedError(app)
+    assert str(err_info.value) == "Cannot create instance of class 'example.application.Application' named 'example', it already exists!"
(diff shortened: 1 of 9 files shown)


ginty:
Looks good.

Wonder if we should namespace these? e.g. OrigenDuplicateError, OrigenNotFoundError, etc. ?

example/tests/errors_test.py
```diff
@@ -0,0 +1,15 @@
+import origen
+import pytest
+from origen.errors import *
+
+def boot_falcon():
+    if origen.app is None:
+        return origen.app.instantiate_dut("dut.falcon")
```
ginty: I think this returns the DUT, or at least, it should!

info-rchitect:
> Looks good.
> Wonder if we should namespace these? e.g. OrigenDuplicateError, OrigenNotFoundError, etc. ?

I would be fine with that, let's see if anyone else has any concerns.

coreyeng:
I think they're already kind of namespaced by being in the `origen` module. importing would end  up looking like `from origen... import OrigenError`, which may be a bit redundant.

coreyeng:
You can always do an `import as ...` if there's conflicts in a particular case as well.

ginty:
Yeah good point, I'm still thinking in Ruby terms I guess.

example/tests/errors_test.py
```diff
@@ -0,0 +1,15 @@
...
+    if origen.app is None:
+        return origen.app.instantiate_dut("dut.falcon")
+    else:
+        return origen.app
+
+def test_duplicate_detected_error():
+    app = boot_falcon()
+    with pytest.raises(DuplicateDetectedError) as err_info:
+        raise DuplicateDetectedError(app)
+    assert str(err_info.value) == "Cannot create instance of class 'example.application.Application' named 'example', it already exists!"
```
ginty: I think we should go back to calling this your original name (DuplicateInstance ?).
The reason I said the more generic name was that I imagined that it could be used for many different duplicate scenarios and the error message would be defined by the specific case.
However, here the error message couples it to a duplicate instance (which is nice that this module takes care of it), so we should call it a more descriptive error name as you originally had it.

info-rchitect: I made the pytest a real world example, let me know if my `sub_blocks` implementation looks correct.

ginty:
Thanks for the update to the error name, I don't see the logic behind the sub_blocks change though (and a shame you coupled it this PR).

ginty:
Sorry, I do see why you did it now, but this is going down a path I don't really like.
Rust is the source of authority on what's in the model and should be bubbling up errors rather than adding an additional layer on top in Py.

ginty:
In fact, it should already be raising an error in this case, does it not?

https://github.com/Origen-SDK/o2/blob/master/rust/origen/src/core/dut.rs#L204

info-rchitect:
> Sorry, I do see why you did it now, but this is going down a path I don't really like.
> Rust is the source of authority on what's in the model and should be bubbling up errors rather than adding an additional layer on top in Py.

If you don't have `sub_blocks` in Python, how can you iterate?

ginty:
I think if you try `dut.sub_blocks` you will find that it does return a dict containing the sub_blocks which you can iterate on.

info-rchitect:
> I think if you try dut.sub_blocks you will find that it does return a dict containing the sub_blocks which you can iterate on.

I got an error, am I missing something wrt the API?

~~~python
    def test_duplicate_instance_error():
        boot_falcon()
        assert list(origen.dut.sub_blocks.keys()) == ['core0', 'core1', 'core2', 'core3']
>       for sub_block in origen.dut.sub_blocks():
E       TypeError: 'Proxy' object is not callable

tests/errors_test.py:11: TypeError
~~~

info-rchitect:
@ginty last commit removes the sub block edits, we can figure out direction in another PR.

PR #61, opened 2020-01-03
stub out of to_python

info-rchitect:
Sorry for all of the churn, I had to re-open with latest master.  Here is the [underlying issue](https://github.com/Origen-SDK/o2/issues/58) driving this PR.

PR #66, opened 2020-01-07
Fix sub-block iteration

ginty:
Quick for for issue with sub-block iteration noted in #61 discussion

### DIFF
--- a/example/tests/sub_blocks_test.py
+++ b/example/tests/sub_blocks_test.py
@@ -1,4 +1,5 @@
 import origen
+import pdb
 
 def test_sub_blocks_can_be_added():
     origen.app.instantiate_dut("dut.falcon")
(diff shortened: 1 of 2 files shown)


info-rchitect:
@ginty  Thanks for the quick PR.  What is the difference between `_dict` and `__dict__`?

info-rchitect:
@ginty will this fix then enable the same pattern of iteration for regs, memory maps, and address blocks?

ginty:
I think there is no difference between _dict and __dict__, but I found a comment somewhere when trying to work out wtf was going on that mentioned that __dict__ was used by Python (to store metadata) so I changed it to avoid clashing with that.

Made no difference though, the missing __iter__ method was the key.

Oh to be back in Ruby land when all something like would take is `class Proxy < Hash`

ginty:
This will have no effect on any other iterators, this is purely for the sub_blocks. All the others are implemented in Rust, but sub-blocks are uniquely done in Python because that's where a lot of the user application code will go.

Issue #69, opened 2020-01-08
Compiler/Templater


info-rchitect:
Hi,

So we got no ERB equivalent in Python so we need to go 3rd party.  Here is what I have found so far:

- [Mustache](mustache.github.io/)
- [Jinja](https://jinja.palletsprojects.com/en/2.10.x/)
- [Mako](https://www.makotemplates.org/)

In general solutions like Mustache would make the compiler much less powerful vs Jinja and Mako that can execute conditional Python.  Perhaps it really just comes down to the templating syntax ease of use and the documentation completeness?  Send your opinions.

info-rchitect:
A nice read [here](https://www.fullstackpython.com/template-engines.html).

info-rchitect:
There is also this [Rust version based on Jinja](https://github.com/djc/askama)

ginty:
I think it has to be done in Python rather than Rust since it is key that the template has access to the application variables (which will be in Python).

To have something on a par with O1 it looks like Jinja2 and Mako are the main contenders.

Personally, I prefer the look of Mako - it looks more equivalent to what we got from ERB. That is, you basically write any Python you like in your templates, whereas Jinja2 seems like more of a language subset.

coreyeng:
From my experience, Jinja2 is a bit more popular. Its the default templating (and built by the team) for [Flask](https://www.palletsprojects.com/p/flask/) (Closest alternative to Ruby's Sinatra) and I've seen it used a lot with Django.

Though, I think both have Mako plugins available, but just from a popularity perspective, I think Jinja2 might have an edge.

FWIW, I've used Jinja2 in my Flask apps before and its great. Never used Mako, but that doesn't mean its not great too.

info-rchitect:
Why couldn't we support both?  The compiler could detect the syntax of the compiled file and route it appropriately.

coreyeng:
We definitely could, but the nature of the question made it seem (to me) like we (as the core team) should agree on some standard so at least the stuff we write all looks similar. If that wasn't the purpose, we should probably discuss that since having half of O2's core templates in Jinja2 and the other half in Mako doesn't seem like a good thing.

[...]

PR #70, opened 2020-01-08
stub-out of compiler

info-rchitect:
Based on [this issue](https://github.com/Origen-SDK/o2/issues/69).  Should be able to demonstrate compiling with mako soon enough,.


ginty:
So far so good.

FYI, as far as file arguments go, the plan is for the CLI to deal with file arguments - so to sanitize/validate them, decompose lists into individual files, etc. 

What this will mean for something like the compiler is that there will be an iterable that you can call, something like `origen.input_files`, that you can iterate on or call `next()` to get the next file in the queue.

This is tracked in #3.
 
That will arrive when it arrives and hack together whatever you need to get going right now, but bear in mind that the eventual plan is to integrate with a common file queue.

info-rchitect:
@ginty how will files be handled with programmatic compiler calls vs CLI?

ginty:
That's up to you I suppose.

info-rchitect:
> That's up to you I suppose.

OK then I think I am on the right track as this implementation should work with both.

info-rchitect:
Where should the output directory default to?

ginty:
I would reference it back to `origen.output_directory` for now, you'll have to add that attribute and set it to `{origen.root}/example/output` for now.
We can later hook that up to something from the application config.

info-rchitect:
Am I on the right track here?

info-rchitect:
@ginty How did you get the DUT loaded into the compiler scope in O1.  As you can see in my implementation the user has to `import origen` explicitly in their template file.

ginty:
Python seems to have the concept of local variables and global variables within the scope of a given module (file).
Here is an Origen method which returns what we should consider to be a standard set of globals available throughout an Origen application: https://github.com/Origen-SDK/o2/blob/978a1de758e37af571d43009911132961bd4525c/python/origen/__init__.py#L35

Here is an exec method being used to invoke a Python file with the specific variable context, maybe you can do something similar here - https://github.com/Origen-SDK/o2/blob/978a1de758e37af571d43009911132961bd4525c/python/origen/__init__.py#L31

info-rchitect:
@ginty Are we good to go for an initial compiler release?

ginty:
Would like to see 'origen' and co auto-loaded before we close this

info-rchitect:
@coreyeng Something is broke wrt loading python libraries, can you check the travis config?

coreyeng:
@info-rchitect, Master has been broken for a while now. The reorg somehow caused that. Not quite sure why it didn't pop up before, but I've fixed it on #79. You can either pull the travis config from there or ignore it for now and it'll be fixed when #79 gets merged.

info-rchitect:
> @info-rchitect, Master has been broken for a while now. The reorg somehow caused that. Not quite sure why it didn't pop up before, but I've fixed it on #79. You can either pull the travis config from there or ignore it for now and it'll be fixed when #79 gets merged.

Thx @coreyeng for the help, I pulled it from your PR.

coreyeng:
I'm trying to remember if there were any other changes needed. I think the 64-bit vs. 32-bit in the dict-like API may need bringing in too. Might be simple to wait for #79's merge if this build fails.

coreyeng:
Actually, I think the above for list-like, which would be new in #79, so nevermind that.

info-rchitect:
@ginty @coreyeng We got 5/8 passing and the fails are not related to the compiler code.  Are we good?

ginty:
Looks like a genuine failure to me that should be fixed - looks like the compiler tests (or the compiler) doesn't handle Windows paths properly.

Can you also please merge latest master into this branch to clean up the diff and remove the file-handler-related diffs.
Thanks!

info-rchitect:
@ginty @coreyeng I noticed this error in the Windows build:

~~~shell
872
Origen: INFO (00:00:00.000): 2.0.0-pre0
873
Is a suitable Python available? ... YES
874
Is the internet accessible?     ... YES
875
Is a suitable Poetry available? ... NO
~~~

Is there an Origen log saying what failed in the `origen setup` command?

ginty:
That's not an un-recoverable failure, what happens after that?
It should go on to install Poetry.

info-rchitect:
OK we clean!

PR #72, opened 2020-01-09
Moved pyapi outside of origen crate

ginty:
I was getting a number of weird build errors like the compiler complaining that some vars should not be mutable when building pyapi, while requiring the opposite when building the core.

This change moves the pyapi/ to its own crate alongside the main origen crate rather than being embedded within it.
Its builds much cleaner now for me and is more of a sensible layout.

Hopefully it is not too difficult to merge existing branch code. The, change is really just to move the pyapi directory up one level and then a few config file updates to account for that.


coreyeng:
Context for errors: https://github.com/Origen-SDK/o2/blob/master/.travis.yml#L23-L25

ginty:
Hi @coreyeng, OK I'll go ahead and merge this if you wouldn't mind having a go at fixing the build.
I had a go and I don't understand why it's failing, the main difference of the .so moving from `rust/origen/target` to `rust/pyapi/target` looks to have been accounted for to me.

`cargo test` should now be working in `rust/origen` for all Pythons, though I don't think it should change anything here since there is no Python dep. `rust/pyapi` however still has the same issue you mention.

Issue #31, opened 2019-12-19
Support enumeratedValues in registers


chrisnappi:
IPXact supports bitfield enumerated values:
https://accellera.org/images/xmlschema/spirit/1685-2009/memoryMap.xsd
search for:
<xs:element name="enumeratedValues">

We should support this, so that we can do things like:
dut.reg(:clock_ctrl).bits(:clock_select).write(:osc_24mhz)

ginty:
Yes this will definitely be done, the data structure for it is already there - https://github.com/Origen-SDK/o2/blob/master/rust/origen/src/core/model/registers.rs#L124

Just need to come up with the appropriate API.

Issue #71, opened 2020-01-09
Database organization


ginty:
Currently the Rust database is organized under one entity, the DUT.

When adding registers recently this proved to make things very tricky when trying to deal with two immutable collections at the same time - adding a new register and the bits for that register.
Ultimately I made it work, but the code could not be organized as I would have wanted it and life would have been much easier if I could get independent locks on the register and bit collections.

So the takeaway is that for the global database collections, finer granularity is better, and I plan to transition away from everything under `DUT`, to having individual collections `REGISTERS`, `BITS`, etc.

Pins may already be doing this anyway, and for anything new that is being added by @coreyeng or @pderouen, standalone static variables should be used.

I don't think that should affect the structure of wider code much at all, just a matter of changing the variable you reference, and in many cases these collections are being treated individually anyway rather than the `DUT` providing some sort of central API - e.g. https://github.com/Origen-SDK/o2/blob/master/rust/origen/src/core/dut.rs#L46

coreyeng:
I think if we can get interior mutability to work it'd be cleaner in the long run. Keeping track of and cleaning up statics may be tricky if we start going multi-threaded. And if we're using static variables, we could introduce the static lifecycle again which would allow us to store references to these static variables instead of using an ID lookup table. Not that the lookup table is inefficient, but it is an added layer that could be removed if we use statics.

PR #73, opened 2020-01-14
Regs

ginty:
There's still a lot to do on regs, but thought I would open a PR to merge the current state into master to allow others access to it.
Work will continue to add more functionality on this branch afterwards - I've copied over the reg_spec from O1 and my plan is to now work through that to add whatever is missing.

This PR adds the following:

- More console displays, including memory maps, address blocks and registers
- Data can be read and written to registers and bit fields

Here is the latest register definition API:

~~~python
    # This is the reg description
    with Reg("adc_ctrl", 0x0024, size=16):
        # This is the COCO description
        Field("coco", offset=7, access="ro")
        Field("aien", offset=6)
        Field("diff", offset=5)
        Field("adch", offset=0, width=5, reset=0x1F, enums={
            # A simple enum
            "val1": 3,
            # A more complex enum, all fields except for value are optional, it will probably
            # eventually support assigning the description from this comment
            "val2": { "value": 5, "usage": "w", "description": "The value of something"},
        }
~~~

Note that enumerated values as discussed in #31 are partially implemented - any enums defined as above do end up in the database but there is no API to consume them yet (will likely be `my_reg.set_data("val1")`).

@info-rchitect I commented out your translator tests due to API changes, please feel free to re-instate with the latest API or wait for things to settle down as you see fit. Hopefully the above reg API is stable now though.

In the console, you should soon be able to just do `dut.regs` to see the list of regs that have been defined as they would have been in O1 without a memory map or address block parent.
However, for now you need to do `dut.default.default.regs` (reflecting that such regs are now physically instantiated in a default (auto-generated) memory map and address block), and you will see something like this:

~~~text
dut
 └── memory_maps['default']
      └── address_blocks['default']
           ├── register_files []
           └── registers
                ├── reg1
                ├── reg2
                ├── areg0
                └── freg1
~~~

And register displays now work, `areg0` can be looked up a number of ways:

~~~python
dut.areg0
dut.reg("areg0")
dut.default.default.areg0
dut.memory_map("default").address_block("default").reg("areg0")
~~~

And you get the same reg view from O1:

~~~text
0x1000 - friendly.path.to.be.implemented.areg0
  ╒═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╕
  │     15      │     14      │     13      │     12      │     11      │     10      │      9      │      8      │
  │             │             │             │             │             │             │             │             │
  │             │             │             │             │             │             │             │             │
  ├─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┤
  │      7      │      6      │      5      │      4      │      3      │      2      │      1      │      0      │
  │    coco     │    aien     │    diff     │                              adch[4:0]                              │
  │    0(RO)    │      0      │      0      │                                 0x0                                 │
  └─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┘
~~~

The data can be get/set on the whole register and individual bit fields:

~~~python
dut.areg0.get_data()     # => 0
# data() is an alias of get_data()
dut.areg0.data()          
dut.areg0.set_data(0xFFFF)
dut.areg0.get_data()     # => 0x7F   (RO and missing bit behavior is implemented)

# This should work soon:
# dut.areg0.adch.data()
# But for now, the longhand works:
dut.areg0.bits("adch").data()    # => 0x1F
dut.areg0.bits("adch").set_data(5)
~~~


info-rchitect:
@ginty Have you hooked up regs, memory maps and address blocks so this iteration pattern can be used in Python?

~~~python
for reg in origen.dut.regs:
~~~

ginty:
@info-rchitect, it will definitely work like that eventually, but not sure if it does currently or not

ginty:
@coreyeng, I haven't made any changes per #71 yet and I'm currently in two minds about whether I should or not.
I think for non-model related things (e.g. pattern, flow AST) should not pile in under the existing DUT, but its actually working pretty well for me now as far as model-data and state are concerned.

The RwLock provides (threadsafe) interior mutability to the Bit objects which have a state field which is wrapped by it.
So as far as regs go, everything else is essentially immutable, i.e. you declare memory maps, regs, etc. and then these are fixed. The only thing that needs to change during pattern generation is the data stored by the regs and that boils down to the individual bit's state fields.

The way I'm writing it now is that every PyApi entry method grabs an immutable reference to the DUT, and then this is passed down to core functions that require it. Pure Rust entry points (i.e. from the CLI) in future should do the same.
Interior functions should be banned from ever grabbing a DUT lock directly (as that can deadlock easily) and instead must look to their caller to provide it.

That has given me pretty minimal bother from the borrow-checker and I think this is following how data is intended to be shared between functions in Rust.
The interior mutability aspect of the bits means that I hardly ever need a mutable DUT reference and that certainly helps with borrowing.

An advantage I'm seeing from having everything under DUT now is that it is just one reference I need to pass around. Even though moving to separate REGS, BITS, statics does give finer-grained locking, it would then mean that I would have to deal with passing lots of these references around instead of just one.

coreyeng:
@ginty, sounds good! I think I've been doing something similar with timing so it should follow your rules above.

PR #79, opened 2020-01-29
Timing2

coreyeng:
Hello,

Here're some updates to give `timesets` a bit more flexibility. The structure was based off the STIL spec and is organized more or less as a tree via `timeset -> wavetable -> wave group -> wave -> event`.

Currently, instantiating timing can be rather verbose. Adding just a couple of waves can get a bit messy.. [see the instantiation here](https://github.com/Origen-SDK/o2/blob/timing2/example/tests/timing_test.py#L242-L311). That said, this PR is more just to get the general structure down. I think that this structure can represent any timesets that we'd encounter and should be able to handle complex timings that we see in scan (and modeled off of STIL for that reason).

For this PR, I'm looking more for feedback on timing that we still can't represent. #63 was just to get something on the board and with this I'm hoping to get an interface that can represent anything we'd need. If there doesn't seem to be any missing pieces here for the representation itself, then I'll start adding some more user-friendly features to it. Context managers, some shortcut methods (`add_drive_wave, add_verify_wave, etc.`), and the like can all be added later as long as the underlying representation is sufficient. Though, crazy timings may still need to fall back to the more verbose API, but I think that'd be expected, and why its there.

The only things I think I'm missing from STIL are inheritance and bases. I did some [minor inheritance here, just a wave-to-wave inheritance,](https://github.com/Origen-SDK/o2/blob/timing2/example/tests/timing_test.py#L359-L366) to make sure that the structure I had in the backend wasn't going to hit the borrow-checker wall. It seems to hold up alright with a wave creation able to access another wave and its events, so moving up the ladder shouldn't be a problem. This'll also help with the above point in addition to rounding out STIL support.

Some other things that're along for the ride:
* [Travis seems to be fixed now so we can hopefully get CI going again](https://travis-ci.org/Origen-SDK/o2/builds/643174051).
* I added a [`list-like`](https://github.com/Origen-SDK/o2/blob/timing2/rust/pyapi/src/meta/py_like_apis/list_like_api.rs) trait and an accompanying [pytest spec](https://github.com/Origen-SDK/o2/blob/timing2/example/tests/shared/python_like_apis.py#L98-L291) to go along with the `dict-like` from before. Adding a `contains` test for backend objects requires overriding the equality operators, which I haven't tried yet, so for now I just have a [basic test to make sure it doesn't crash](https://github.com/Origen-SDK/o2/blob/timing2/example/tests/shared/python_like_apis.py#L249-L255). For non-Origen objects though, it should work alright.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,13 @@ rust:
 # For the backend, only a single Python version is needed.
 # For the frontend, it'll be tested against the Python versions below, corresponding to the latest Chocolately release for 3.6, 3.7, and 3.8.
 # Additional Python versions, or even frontend lanuages, can be added here as needed.
+# Note: I'm not sure why the LIBPYTHON doesn't necessarily match the Python version. But, going with what gets installed 
+# -
 env:
-  - O2_REGRESSION=BACKEND O2_PYTHON_VERSION=3.8.1 LINUX_BIN_PATH=3.8
-  - O2_REGRESSION=FRONTEND O2_PYTHON_VERSION=3.8.1 LINUX_BIN_PATH=3.8
-  - O2_REGRESSION=FRONTEND O2_PYTHON_VERSION=3.7.6 LINUX_BIN_PATH=3.7
-  - O2_REGRESSION=FRONTEND O2_PYTHON_VERSION=3.6.8 LINUX_BIN_PATH=3.6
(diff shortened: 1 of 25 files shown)


coreyeng:
@Origen-SDK/o2-core-developers Any objections to merging this (after I resolve the conflicts)? I've got a prototype of the tester which I built off of this. I can either just push it to this branch, or merge this and open a new PR is a few days for the tester.

Thanks!

PR #80, opened 2020-01-31
Added initial file argument handler

ginty:
Adds initial handling of the CLI command FILE argument.

@info-rchitect, here's the point for you to hook in the template compiler - https://github.com/Origen-SDK/o2/blob/compile_cmd/python/origen/boot.py#L114

Eventually the Rust side of this will be enhanced to expand directory and list file references, and maybe even handle some basic ifdef-style logic within list files. Nothing is expected to change on the Python side.

Invoke from the CLI as you would expect: `origen c path/to/template.txt`

### DIFF
--- a/python/origen/boot.py
+++ b/python/origen/boot.py
@@ -2,6 +2,7 @@
 from __future__ import print_function, unicode_literals, absolute_import
 
 import sys
+import pdb
 
 if sys.platform == "win32":
     # The below is needed only for pyreadline, which is needed only for Windows support.
@@ -88,7 +89,7 @@ if sys.platform == "win32":
 
 # Called by the Origen CLI to boot the Origen Python env, not for application use
 # Any target/env overrides given to the command line will be passed in here
-def __origen__(command, target=None, environment=None, mode=None):
+def __origen__(command, target=None, environment=None, mode=None, files=None):
     import origen
     import _origen
     import origen.application
@@ -99,10 +100,18 @@ def __origen__(command, target=None, environment=None, mode=None):
     else:
         origen.set_mode(mode)
 
+    if files is not None:
+        _origen.file_handler().init(files)
+
     origen.target.load(target=target, environment=environment)
 
     if command == "generate":
         print("Generate command called!")
+
+    elif command == "compile":
+        for file in _origen.file_handler():
+            # Invoke compiler here
+            pass
 
     elif command == "interactive":
         import atexit, os, sys, colorama, termcolor, readline, rlcompleter
--- a/python/origen/boot.py
+++ b/python/origen/boot.py
@@ -2,7 +2,6 @@
 from __future__ import print_function, unicode_literals, absolute_import
 
 import sys
-import pdb
 
 if sys.platform == "win32":
     # The below is needed only for pyreadline, which is needed only for Windows support.
--- a/python/origen/target.py
+++ b/python/origen/target.py
@@ -6,6 +6,7 @@ def load(target=None, environment=None):
     app = origen.app
     if target == None:
         target = _origen.app_config()["target"]
+
     if environment == None:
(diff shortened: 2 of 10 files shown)


info-rchitect:
@ginty Did you settle on a default templates dir?

info-rchitect:
@ginty The file handler does not return a file variable.

~~~ruby
    elif command == "compile":
        for file in _origen.file_handler():
            breakpoint()
            pathlib.Path(file)
            # Invoke compiler here
            pass
~~~

The breakpoint does not get triggered when I run this command:

~~~shell
origen c ../python/templates/dut_info.txt.mako
~~~

ginty:
It does work, did you compile Rust after checking out?

ginty:
> @ginty Did you settle on a default templates dir?

I think it should be `<app>/templates`, so `example/example/templates` in this case.

info-rchitect:
> It does work, did you compile Rust after checking out?

Yes I built both the origen and pyapi using cargo

info-rchitect:
BTW I merged this branch with my stub-out compiler branch and then did the compile

ginty:
Just tried it and it works fine, maybe the CLI has not been re-compiled?

I think the README could be made clearer on this, but these are the steps to compile Rust:

To compile the Python lib:

~~~
cd rust/pyapi
cargo build
~~~
 
To compile the CLI (not needed often, but it is here):

~~~
cd rust/origen
cargo build --workspace --bins
~~~

If it still doesn't work, then is it something to do with the breakpoint() method?
I've never used that so not sure what it should do, but I'm just testing it with `print(file)` and I can see the file args as I would expect.

info-rchitect:
> Just tried it and it works fine, maybe the CLI has not been re-compiled?
> I think the README could be made clearer on this, but these are the steps to compile Rust:
> To compile the Python lib:
> cd rust/pyapi
> cargo build
> 
> To compile the CLI (not needed often, but it is here):
> cd rust/origen
> cargo build --workspace --bins
> 
> If it still doesn't work, then is it something to do with the breakpoint() method?
> I've never used that so not sure what it should do, but I'm just testing it with print(file) and I can see the file args as I would expect.

Re-did the compilation and no luck.  Using `print(file)` did not work either.  Will try using your branch.

info-rchitect:
@ginty I checked out the `compile_cmd` branch and ran the `c` command with this debug code:

~~~python
    elif command == "compile":
        print("found compiler hook-up")
        for file in _origen.file_handler():
            print(file)
            # Invoke compiler here
            pass
~~~

The command outputs the following to the terminal:

~~~shell
found compiler hook-up
~~~

Can you try the same thing?

ginty:
I just tried it a few mins ago, it definitely works.
I still think you are not running the updated CLI, check which 'origen' you are pointing to and verify the data/time on it.

info-rchitect:
> I just tried it a few mins ago, it definitely works.
> I still think you are not running the updated CLI, check which 'origen' you are pointing to and verify the data/time on it.

yep I am an idiot

ginty:
Cool, glad you got it resolved.
I think it does show that we need to have the CLI passing in a version number to Python and then the Python side can verify against some minimum CLI version.

info-rchitect:
~~~shell
$ origen c ../python/templates/dut_info.txt.mako
Origen: INFO (00:00:00.206): Compiling mako template /mnt/c/o2/compiler/python/templates/dut_info.txt.mako
Origen: INFO (00:00:00.210): Compiler output created at /mnt/c/o2/compiler/python/templates/dut_info.txt
$ ls -alrt  /mnt/c/o2/compiler/python/templates/dut_info.txt
-rwxrwxrwx 1 me me
 33 Feb  6 10:18 /mnt/c/o2/compiler/python/templates/dut_info.txt
~~~

PR #82, opened 2020-02-10
Latest regs

ginty:
Just a periodic merge to get the latest regs APIs into master since @info-rchitect is planning to use them for his regs importer.

If you look at the o1_regs_test.py you will see the stuff that has been added and that is currently about 50% of the way through implementing the APIs from O1.


ginty:
I'll check the tests are passing before merging this, don't have my dev env available right now to confirm.

Issue #97, opened 2020-03-31
Remote app execution mode


ginty:
As you know in O1 Origen has two ways of working:

- Globally, outside of an app workspace
- App mode, inside of an app workspace

It has already been discussed elsewhere that for O2 we will probably add a 3rd:

- Origen development, which would open up some commands related to building/deploying/testing Origen itself (more useful in O2 due to the multi-lang env)

And now I would like to discuss the addition of a 4th:

- Remote app mode

In this mode Origen would act like it was inside a particular application's workspace but while it is actually in some other directory outside of the application.

The relationship between the CLI and the application would be established by one of two ways:

- A file in the current directory or its ancestors, e.g. `.origen_app` which is either a sym link to the application directory or a file containing the path to it
- An environment variable, ORIGEN_APP_DIR

I think order of precedence for establishing what the current app is would probably be:
Being in an app workspace -> ORIGEN_APP_DIR -> .origen_app  and in some cases probably a warning would be emitted (e.g. when a .origen_app is found but is being overruled by the ENV variable).

The intended use case for this is when Origen is being used as a platform for creating commands and supporting functionality within a wider 3rd party tool environment.
e.g. When it is used in design/verification environments where an Origen application is created and managed by one or two engs, it is then checked in to the wider environment and then all users get access to custom origen commands without really caring or being aware of how they were built and made available.
@chrishume, thinking about your use cases here.

As part of this it will mean that Origen execution will need to be well de-coupled from whatever is the pwd, so it will also resolve weaknesses in O1 where some (many?) commands didn't work properly unless the pwd was the app's root dir.

Will probably start on this next week, so keen to hear any comments/suggestions/objections before then.

Thanks!

coreyeng:
While were talking about the CLI, do you think it will remain in Rust long term? In O1, we had the ability to add/extend commands from the application or (eventually) plugins which isn't really possible considering the CLI is now compiled. We could wrap in some interface into the Rust side, but then we'll need to fill in a lot of the things `clap` does for us, like the command organization, and `help` dialog.

I wonder if it might be better to move the CLI to Python, which can dynamically build up commands and then have a bunch of helpers on the O2 side to decouple it as much as possible from the frontend. So hings like `origen t` will start on the Python side, but then immediately go into passing the command back to Rust, which would amount to running the same `target` code that is run today, just the starting point is in the frontend instead.

[...]

PR #100, opened 2020-04-22
Command Launcher Update

### LINKED ISSUE #27: Bug when used for paths with spaces
This code does not work when the command contains a path with spaces:

https://github.com/Origen-SDK/o2/blob/c4662b5fb677b4e7d265e71e89f0a17fbd9687ab/rust/origen/src/core/os.rs#L24-L26

According to this comment using <code>cmd /C command_to_execute</code> argument is "normally" not needed. But, I suppose we have a case where it is needed.
https://stackoverflow.com/questions/44757893/cmd-c-doesnt-work-in-rust-when-command-includes-spaces

If I construct the return value exactly the same way as it's done for Linux, I don't have any issue. What led to the need for executing like this on Windows?

pderouen:
fixes #27 

@ginty recommended using <code>PathBuf</code> it cleaned up a lot of the issues I was seeing. I don't know that all of the added <code>OsString</code> is necessary. In my testing I had better luck when using <code>OsString</code>. In any case the string has to be converted to the appropriate type for the system eventually. So, no harm (in my view).

@coreyeng can you kick the tires a bit?

I think the os.rs file becomes basically unused in this branch. I left it in for now.

### DIFF
--- a/rust/origen/cli/src/commands/setup.rs
+++ b/rust/origen/cli/src/commands/setup.rs
@@ -2,11 +2,13 @@ extern crate time;
 
 use crate::python::{poetry_version, MIN_PYTHON_VERSION, PYTHON_CONFIG};
 use online::online;
-use origen::core::os;
+//use origen::core::os;
 use origen::core::term::*;
 use std::fs;
 use std::io;
 use std::path::PathBuf;
+use std::ffi::OsString;
+use std::process::Command;
 
 const POETRY_INSTALLER: &str =
     "https://raw.githubusercontent.com/sdispater/poetry/1.0.0b7/get-poetry.py";
@@ -64,13 +66,15 @@ pub fn run() {
                 fs::read_to_string(&get_poetry_file).expect("Unable to read Poetry install file");
             let new_data = data.replace(
                 "/bin/env python",
-                &format!("/bin/env {}", PYTHON_CONFIG.command),
+                &format!("/bin/env {}", PYTHON_CONFIG.command.clone().into_string().unwrap()),
             );
             fs::write(&get_poetry_file, new_data).expect("Unable to write Poetry install file");
 
             // Install Poetry
-            os::cmd(&PYTHON_CONFIG.command)
-                .arg(format!("{}", get_poetry_file.display()))
+            //os::cmd(&PYTHON_CONFIG.command)
+            Command::new(&PYTHON_CONFIG.command)
+                //.arg(format!("{}", get_poetry_file.display()))
+                .arg(get_poetry_file)
                 .arg("--yes")
                 .status()
                 .expect("Something went wrong install Poetry");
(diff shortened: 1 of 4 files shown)


coreyeng:
Thanks! This is working okay for me so far. If we're not using `os.rs` anymore then that can be removed.  The command handler there was really just to get something going on Windows.

pderouen:
Sounds good. I will remove os.rs.

pderouen:
<code>on_windows</code> and <code>on_linux</code> are being used by pyapi. So, I cleaned out the command launcher code instead of deleting os.rs.

pderouen:
pushed an update that fixes #101 

Do the Travis builds test the CLI commands?

ginty:
> Do the Travis builds test the CLI commands?

No, not yet

rust/origen/cli/src/commands/setup.rs
```diff
@@ -64,23 +65,23 @@ pub fn run() {
...
 
             if poetry_version().unwrap().major != 1 {
                 // Have to use --preview here to get a 1.0.0 pre version, can only use versions for
                 // official releases
-                os::cmd(&PYTHON_CONFIG.poetry_command)
-                    .arg("self:update")
-                    .arg("--preview")
+                Command::new(&PYTHON_CONFIG.poetry_command)
+                    .arg(OsString::from("self:update"))
+                    .arg(OsString::from("--preview"))
```
ginty, line 84: Is this OsString creation really required here?
I think we have a lot of other places where we don't do this. We should either do this everywhere or nowhere I think.

ginty:
If you wanted to add some CLI tests I would add a `cli_test.py` here - https://github.com/Origen-SDK/o2/tree/master/example/tests

You should be able to switch dirs out of the app directory in Python code if you wanted to test global commands too.

Note that I think the naming of `_test.py` in the file is important, as is naming the test functions `test_..`.

To invoke those tests run `poetry run pytest` for them all, or `poetry run pytest tests/cli_test.py`.
I think there are also ways to invoke specific tests only, though I'm not sure of the details.
The test framework is called pytest if you want to google it.

rust/origen/cli/src/commands/setup.rs
pderouen: No, I'm not sure that it is needed here. The arg should be internally converting to OsString. I will back out the OsString changes and give it another try.

pderouen:
> If you wanted to add some CLI tests I would add a `cli_test.py` here - https://github.com/Origen-SDK/o2/tree/master/example/tests

It was just an honest question, wondering if the linux compile of the cli gets the tires kicked. I don't know that it's necessary to directly test the cli. As we're working on updates it'll be getting a constant run through. I'm willing to take a stab at a few tests to at least check that it doesn't die on <code>origen v</code> or <code>origen g</code> if you'd like to see that.

coreyeng:
Partially from #97 , and partially from the mess I've got [here](https://github.com/Origen-SDK/o2/blob/doc_system/rust/origen/cli/src/bin.rs#L326-L356), I think we'll need to rethink the command handler a bit before its all said and done. Might not be a bad idea to have some CLI tests to go off of. Not that it needs to be done now, just thinking out loud.

pderouen:
Yeah, I think I changed my mind about tests for the CLI over night. It would be better to have tests for it. I'll put a little thought into it over the next day or 2, but I think I'd rather merge these changes as is.

pderouen:
This seems like a good way to test the cli:
https://rust-cli.github.io/book/tutorial/testing.html#testing-cli-applications-by-running-them

pderouen:
@coreyeng part of adding testing I was thinking to add for the cli _might_ involve moving the "do something with the args" code into a lib for direct unit testing... so instead of checking that <code>origen g</code> works when with a windows style path argument; I would just check that the strings passed to poetry are literals (and make that more likely by using a central fn or macro to build them).

Anyway, all of that to ask if you'd want to wait until after your branch is merged and I can work from there to add the testing? Or, what?

pderouen:
I'm starting a cli_test branch to work on tests. I'll go ahead and merge this.

PR #103, opened 2020-04-27
Add README.md helpers

pderouen:
Code organization comments.


ginty:
Thanks for driving this @pderouen, great to see you back!

I think mini-guides as to the big picture like you have started here are what we should focus on right now rather than coupling too much to absolute file and function references which are still in a state of flux in many areas.

In that spirit I have added some more detail to the guide on how Origen boots.

Let me know what other topics you would like similar details on and I'll do my best to fulfil them in a timely manner.

pderouen:
Thanks for the extra details.

PR #110, opened 2020-05-26
Added fmt and update commands

ginty:
Adds an `origen fmt` command which will code format (similar to O1 lint command) all Python files and when in an o2 development workspace it will do all Python and Rust files.

Will probably add app config options in future to allow apps to configure files/dirs to exclude, but for now it just does everything.

`origen::STATUS` can also now be interogated to determine if Origen is executing within an o2 workspace or not.

I added an additional command `origen update` which is to be used to update an app's Python dependencies, similar to running `bundle update` in O1. This is just a thin wrapper around the underlying poetry command.

This fmt tool requires a new Python package, so to run it in an existing workspace you must first run `origen update` within the example app.

I haven't run the tool on this source code yet because it will introduce a lot of diffs in the Python code.
Once this is merged I will run it on master and that will give folks an easy-ish path to merge into their branches:

* Checkout the version of master that includes this PR
* Merge that into their branch
* Run `origen fmt` on their branch
* Checkout the latest version of master that has the formatting applied
* Merge that into their branch

### DIFF
--- a/.origen_dev_workspace
+++ b/.origen_dev_workspace
@@ -1,0 +1,1 @@
+This file simply lets the CLI know that it is in an Origen development workspace
--- a/example/poetry.lock
+++ b/example/poetry.lock
@@ -4,7 +4,7 @@ description = "Atomic file writes."
 name = "atomicwrites"
 optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
-version = "1.3.0"
+version = "1.4.0"
 
 [[package]]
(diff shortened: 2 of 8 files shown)


ginty:
Actually, just running `origen setup` in an existing workspace with this update might be sufficient. I think `origen update` is only required when you add/change pyproject.toml and want to update the lock file accordingly.

ginty:
Thanks @pderouen, @coreyeng any objections to merging this?

pderouen:
I don't see any issue with merging it.

PR #111, opened 2020-06-09
Proj command updates

ginty:
Some updates to the proj command:

* Various new sub-commands for viewing modifications, undoing local mods and tagging.
* Corresponding updates to the Git and DesignSync drivers, both of which are now in good shape and fully implement - initial population, checkout of a new version (with both merge and overwrite of local mods supported), report status (new, changed, deleted files), rollback local mods, and tag.
* Added the ability to define groups of packages in the BOM to assign a single version and/or otherwise work with a collection of packages as a single unit.
* Packages from Git can be defined with an ssh and https url, with the former being tried first and falling back to https if it doesn't work.
* Current user name and email will be fetched from Git unless otherwise defined.
* Added some tests for it.


ginty:
Sorry, just going ahead with this merge as I'd like to tag and install it in NXP.
I doubt that it should be of much concern though as it is a pretty self-contained feature and nothing is set in stone at this point anyway.

PR #112, opened 2020-06-10
Fix user ID fetch for Linux

ginty:
When releasing an initial version of the proj command it failed miserably because the function to determine the user ID didn't work.
It did work in a debug build, but not in a release build.
The root cause is inside a whoami crate that I used and which seems to use some clib internally - i.e. not easy to debug and I would like to get rid of it.

This update transitions to a less magic approach of simply running 'whoami' and capturing the output, and it now works for the production release Linux build.

@pderouen, any chance you could look at the Windows branch of this?
Is there a similar command we could run in Windows space?

Note that one thing I discovered here is that `if cfg!(linux)` does not work, it has to be either `if cfg!(unix)` or `if cfg!(target_os="linux")`.

### DIFF
--- a/rust/origen/Cargo.lock
+++ b/rust/origen/Cargo.lock
@@ -1443,7 +1443,7 @@ dependencies = [
 
 [[package]]
 name = "origen"
-version = "2.0.0-pre0"
+version = "2.0.0-pre1"
 dependencies = [
  "built",
  "chrono",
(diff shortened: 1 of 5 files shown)


ginty:
@pderouen, are you planning to push an update for the Windows path here, or will we just go ahead and merge this?

pderouen:
This branch works as is from Windows

PR #118, opened 2020-08-14
Doc system tweaks

ginty:
Various tweaks, mainly for Python 3.6 compatibility I think, to get the doc system running on my setup

### DIFF
--- a/python/origen/target.py
+++ b/python/origen/target.py
@@ -30,7 +30,8 @@ def load(targets=None):
             if callable(t):
                 t()
             else:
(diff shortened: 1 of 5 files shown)


ginty:
This doesn't totally fix it on a Linux system I don't think, see here for details - https://github.com/Origen-SDK/o2/issues/119

coreyeng:
This is working for me in Windows with 3.8. If you want to merge this, I'll pull down a new master and look at #119 and whatever seems to be making `subproject` unhappy (happens in master too, so probably something leftover from my previous environment).

Issue #119, opened 2020-08-14
Doc includes not working


ginty:
Hi @coreyeng,

When generating either the O2 guides or the web site for an app I get errors like this due to the common shorthand includes not being found:

~~~
/home/stephen/Code/github/o2/test_apps/python_app/web/source/index.rst:1: WARNING: Problems with "include" directive path:
InputError: [Errno 2] No such file or directory: 'web/source/home/stephen/Code/github/o2/test_apps/python_app/web/source/interbuild/shorthand/origen.rst'.
~~~

For the O2 guides this ultimately leads to a failure to generate because various shorthand references end up undefined.

I've verified that the include directive is being generated correctly:

~~~
.. include:: /home/stephen/Code/github/o2/test_apps/python_app/web/source/interbuild/shorthand/origen.rst
~~~

I think the problem is that this include directive doesn't accept a true absolute path, rather paths starting with `/` are interpretted as a reference from the top source directory, not the top of the filesystem.
I wonder if it does the right thing on Windows when faced with something like `C:/` because it is un-ambiguous in that case.

Anyway, I couldn't find any way to make it respect a true absolute path and was thinking of going down the route of copying the files from Origen core into the local app's source folder at the beginning of generation and thereby enabling a relative path to be used regardless of where Origen is installed.

However, I guess it will be more efficient if you look at a fix since your more aware of what all of the similarly affected resources here could be.

Thanks!

PR #120, opened 2020-08-14
New app

ginty:
Adds an 'origen new' command for building a new application and for generating content (e.g. a new dut) within an application.

The generated content could be better and will evovle over time, however this should establish the infrastructure for both features allowing further expantsion by the community.

All templates reside in `rust/origen/cli/src/commands/new/templates`.

In the case of the new app it is simply a case of modifying those file and/or addition additional files into the template directory and they will be picked up.


coreyeng:
I think I got the web stuff working on Linux now. Python 3.6 throws a warning as it doesn't like the [fanciness in making an abstract class method a property](https://github.com/Origen-SDK/o2/blob/master/python/origen/compiler.py#L27-L30), but I envision we'll eventually tie the website release to Travis and then we can spin the docs with whatever setup we want. It seems fine in 3.8, though it could also be re-written without much effort since the compiler is still small.

This should take of #119 too.

ginty:
Hi @coreyeng,

> Back to the new app though, I also ran into a small issue with running from a local origen build. It'll find the origen module, but it doesn't find the _origen library in that directory. I worked around it by setting the PYTHONPATH in the app though. Not sure if you've tried this. Could just be a Windows thing too. I'll keep looking at it. @pderouen, this may affect you as well.

I noticed this too and was plannig to fix it on another branch where I'm setting up the Python plugin system. FYI the changes I was planning to make there are as follows:

- Consolidate the existing `origen setup` and `origen update` commands into a single command, `origen env`, such that these will now be `origen env setup` and `origen env update`
- Add a new switch to configure an app to use a local origen: `origen setup --origen path/to/my/o2`

When the app is configured for Origen development the `origen build` command will become available within the app and it will work the same way as in an o2 workspace - i.e. just run it from within the app workspace and it will compile `_origen` and link it into the app.

That should avoid any need to mess around with `PYTHONPATH`.

Let me know if you have any comments/requests on this.

Thanks!

ginty:
Hi @coreyeng,

Thanks for the update here.

Regarding this item:

> I think I got the web stuff working on Linux now. Python 3.6 throws a warning as it doesn't like the fanciness in making an abstract class method a property, but I envision we'll eventually tie the website release to Travis and then we can spin the docs with whatever setup we want. 

Can we make it 3.6 compatible?

Understood that as far as the O2 docs are concerned then limiting to 3.8 might not be that bad, but for app docs and for users wanted to build the o2 docs locally (i.e. like me!), I think we should be supporting 3.6.
I still think we should be supporting 3.6 as a minimum version since that is default on Ubuntu/WSL2.

Thanks!

coreyeng:
Hi @ginty,

What you have planned for environment sounds good! I wasn't sure if it was user error, windows problems, or something else. Setting the `PYTHONPATH` as a workaround for now is fine for me though.

I added a workaround for Python 3.6. I have too much hubris and was quite giddy when I got the `abstractclassmethod` to work as a property like I wanted. It works functionally (all the specs pass) and I'll be damned if 3.6's docstrings are what messes it up! :p Instead, I added an event which may come in handy later which allows preprocessing of the docstrings. I just override it there.

Thanks,

Corey

PR #128, opened 2020-11-01
Jtag prototype

coreyeng:
Hello,

I've added a [JTAG](https://github.com/Origen-SDK/o2/blob/b4609b2a945adb57216f01654eb4e7abf6c1580f/rust/origen/src/services/jtag/service.rs) implementation which follows the same structure as [SWD](https://github.com/Origen-SDK/o2/blob/b4609b2a945adb57216f01654eb4e7abf6c1580f/rust/origen/src/services/swd/service.rs). I also added a [JTAG DP](https://github.com/Origen-SDK/o2/blob/b4609b2a945adb57216f01654eb4e7abf6c1580f/rust/origen/src/services/arm_debug/jtag_dp.rs) into our arm debug driver. **The features themselves are limited.** I'm not yet going for full driver support, but this PR takes us from "no-JTAG" to something we can build off of. It is definitely raw, and some things that may make @pderouen cringe, but its better than our current state I think.

This was tested on an internal device I'm working on here, so its actually been verified. I don't have specs, nor do I intend to do them quite yet, as I want to flush out this, swd, arm debug, and an internal driver I'm working, and see how things look at end. I think things are still a bit too fluid to want to make specs, examples, and documentation just yet, but they'll be coming (one of these "fluid things" being `Capture Nodes` as discussed in #125, and actually `capturing` in general, as well as `overlays` which are both unsupported right now)

I've also got a multi-pin internal driver working which is implemented as a service  purely in Python (using the [python plugin stuff](#123 )), so there's some changes in there for that, but nothing that I think is controversial. The only thing I think I actually changed is allowing a default `width` in registers which will resolve to 32-bits with the defaults.

https://github.com/Origen-SDK/o2/blob/b4609b2a945adb57216f01654eb4e7abf6c1580f/rust/origen/src/core/model/registers/register.rs#L299-L304

One thing to note is I spun the [Transaction](https://github.com/Origen-SDK/o2/blob/b4609b2a945adb57216f01654eb4e7abf6c1580f/rust/origen/src/generator/utility/transaction.rs) struct that I use in driver to the [Python side](https://github.com/Origen-SDK/o2/blob/b4609b2a945adb57216f01654eb4e7abf6c1580f/rust/pyapi/src/utility/transaction.rs) as well. Like the Rust-side counterpart, I want to use this as generic means to handle "incoming transactions". I plan to expand the `pins` to be accept this type as well, but that's not in this PR. This is used in my internal driver, but an example of generating one, which denotes a `write transaction`:

image

and allows you to do things like:

image
 
or 

image

without needing to worry about what the input looks like. Its analogous to a lot of the `clean addr` or `clean data` stuff we had floating around the various O1 drivers, just in a more universal space.


rust/origen/src/generator/nodes.rs
```diff
@@ -67,24 +66,27 @@ pub enum Attrs {
...
-        Option<String>,
-    ), // size, data, verify_enable, capture_enable, overlay_enable, overlay_str
-    JTAGWriteDR(u32, BigUint, Option<BigUint>, Option<String>), // size, data, overlay_enable, overlay_str
+        Id, // JTAG ID
+        Transaction,
+        Metadata,
+    ),
+    JTAGWriteDR(
+        Id, // JTAG ID
+        Transaction,
```
pderouen, line 81: Do these Transaction and Metadata structures provide hooks needed to perform a read while write operation?

rust/origen/src/services/jtag/service.rs
```diff
@@ -2,131 +2,306 @@
...
+    // }
+
+    pub fn write_ir(&self, dut: &Dut, transaction: &Transaction) -> Result<()> {
+        let tms = PinCollection::from_group(&dut, &self.tms.0, self.tms.1)?;
+        let tdi = PinCollection::from_group(&dut, &self.tdi.0, self.tdi.1)?;
+        let trans = node!(JTAGWriteIR, self.id, transaction.clone(), None);
+        let n_id = TEST.push_and_open(trans.clone());
+
+        self.comment("Move to Shift-IR");
+        tms.drive_low().cycle();
```
pderouen, line 177: Probably can eventually remove this tms low cycle. It's common practice to park the driver in run test/idle.

rust/origen/src/services/jtag/service.rs
```diff
@@ -2,131 +2,306 @@
...
+
+    // pub fn move_to_state(&self, state: JTAGState) -> Result<()> {
+    //     // ...
+    // }
+
+    // fn _move_to_state(&self, state: JTAGState, tms: &PinCollection) -> Result<()> {
+    //     // ...
+    // }
+
+    pub fn write_ir(&self, dut: &Dut, transaction: &Transaction) -> Result<()> {
```
pderouen, line 170: I guess it'll be the transaction data structure that needs to hold any compare data for a read while write operation

pderouen:
@coreyeng no cringing from me... maybe my standards are lower than you imagined :)

Is this implementation able to handle overlay/capture for 1 or more groups of bits?

rust/origen/src/services/jtag/service.rs
coreyeng: I don't think so. I think it'd be driver specific for how it wants to handle it. A transaction itself is on a single bus, with a single data value, enables, etc. I think it'd be up to the driver accept multiple transactions and handle them accordingly. You can do what you're asking on the Rust side by getting the nodes directly for each transaction and combining them yourself. I can add some helpers for this in the future but I think that'd be a different struct altogether (`ParallelTransaction`, or something like that). It'd get kinda messy making everything in a transaction a vector then having to ensure that vector lengths are equivalent.

Can this be done in O1? I looked at the driver and everything is taking a singular `reg_or_data`.

coreyeng:
Currently nothing handles overlays/captures! That's the next item up in "protocols & subblocks" work.

pderouen:
I more meant to ask if the data structures provide the capability to pass the info. I suppose that would come from the register whose ID is passed in the Transaction. I think the only thing that may be missing is a way to specify both read and write information in a single transaction.

coreyeng:
That would be on two different pins though. I would rather think of it as two different transactions run in parallel. At least in O1, I'm not sure how this is handled, or if its even possible, as a register there doesn't hold both "read data" and "write data", nor does it here either (it resolves to "bits" which hold just a single state).

You can do it with enables if you wanted to read a few bits on TDO, then input some data on TDI (for handling the previous acknowledge, if that's what you're referring to here), but in terms of driving a full data word on TDO while comparing one on TDI, I think that should be thought of as two transactions run in parallel that the driver supports (obviously something like SWD wouldn't support this).

rust/origen/src/services/arm_debug/jtag_dp.rs
```diff
@@ -0,0 +1,300 @@
...
+        } else {
+            self.update_ir(&dut, &services, self.apacc_select as u32)?;
+        }
+
+        let mut t = transaction.clone();
+        t.prepend_data(BigUint::from(0 as u8), 3)?;
+        t.bit_enable = t.bit_enable & BigUint::from(0x7_FFFF_FFF8 as u64);
+        let jtag_service = services.get_service(self.jtag_id)?;
+        let jtag = jtag_service.as_jtag()?;
+        jtag.verify_dr(&dut, &t)?;
```
pderouen, line 249: The ARM debug interface AS specifies a 35 bit write of 0x7 to perform the read operation of DRW and the subsequent shift out of the register data. We've been getting away with shifting in all 0's, hasn't caused any issues.

coreyeng:
Here's how the current `Bit` struct looks:

https://github.com/Origen-SDK/o2/blob/master/rust/origen/src/core/model/registers/bit.rs#L19-L29

We can maybe re-purpose the `device_state`, but I don't think that was its intention.

rust/origen/src/services/arm_debug/jtag_dp.rs
coreyeng: I referenced both, but when I ran into errors, I deferred back to the O1 jtag driver, as I could generate waves with that to actually look for differences in the state.

For that particular line though, that's just setting the enables, not the data. Its "don't-caring" the previous acknowledgment.

pderouen:
> That would be on two different pins though. I would rather think of it as two different transactions run in parallel. At least in O1, I'm not sure how this is handled, or if its even possible, as a register there doesn't hold both "read data" and "write data", nor does it here either (it resolves to "bits" which hold just a single state).
> 

In O1 you passed in a bit collection to the options hash to specify the read data when a write is happening.

It's actually all 1 transaction. JTAG registers are shift registers. As you shift in the LSB, the data currently held in the register's MSB is shifted out on TDO. I wouldn't want the driver to automatically care the out-going bits.

But, there does need to be a way specify it when we need to care TDO during a "write" or specify the data for TDI during a "read". Maybe an optional second Transaction passed in?

rust/origen/src/services/jtag/service.rs
pderouen: I think  your proposal (let the driver accept and handle multiple 1-way transactions) is the best. This is what it looks like to read while write in the o1 driver: https://github.com/Origen-SDK/origen_jtag/blob/bc4dceac90a1a560b1ebf1da2956004a69e70dc1/pattern/rww_test.rb#L13

coreyeng:
I think we need to keep "jtag transaction" vs. "transaction struct" nomenclature separate.

Its one transaction from the DR's perspective on Silicon, but two from the `Transaction` struct's in the code. Or, at least the current implementation of the struct. A transaction struct is meant to represent a single "do this operation on these pins with this data". I think making "operation" and "data" plural in that will get to be messy, partially because a register only holds one state at time, so you'd be "resolving to a transaction" twice. Which is what we'd be doing in O1 if you give jtag two bit collections, which I did [find where that happens](https://github.com/Origen-SDK/origen_jtag/blob/master/lib/origen_jtag/driver.rb#L195-L198). I don't think the arm debug driver uses that, which is my main JTAG usage.

I think it make more sense to either give JTAG two transaction, one for read, one for write, and/or make another struct to represent multiple single transactions happening at the same time. Which, even though the DR is a single register, it amounts to two simultaneous transaction in the vectors.

pderouen:
> I think we need to keep "jtag transaction" vs. "transaction struct" nomenclature separate.

Yeah.
 
> I think it make more sense to either give JTAG two transaction, one for read, one for write, and/or make another struct to represent multiple single transactions happening at the same time. Which, even though the DR is a single register, it amounts to two simultaneous transaction in the vectors.

Still agree that two Transaction structs with 1 being optional is the right choice

coreyeng:
I opened #129 to track this. I think adding a way to do arbitrary parallel transactions would be better than making a one-off case for JTAG, but that's more involved than I wanted to take on in this PR.

pderouen:
Ok, so you prefer to update the Transaction struct to allow parallel Transactions?

I'm fine with merging the existing code either way.

coreyeng:
I'm thinking more of keeping the existing transaction the same but having another struct that is basically a vector of `transactions` and having some functions explicitly geared towards handling parallel transactions.

But, who knows. I may start down that path and that turns out to be worse. Will need to just start playing with it and see what it looks like.

pderouen (approved):
Looks good to me

rust/pyapi/src/services/jtag.rs
```diff
@@ -0,0 +1,185 @@
...
+    fn reset(&self) -> PyResult<()> {
+        let dut = origen::dut();
+        let services = origen::services();
+        let jtag = services.get_as_jtag(self.id)?;
+        jtag.reset(&dut)?;
+        Ok(())
+    }
+
+    #[args(write_opts = "**")]
+    fn write_register(&self, bits: &PyAny, write_opts: Option<&PyDict>) -> PyResult<()> {
```
ginty, line 55: What does write_register do in the context of a JTAG driver?
Would this not be better called `shift_register`?
Generally I think the `write_register` / `read_register` API methods should be reserved for register-map-level drivers.

ginty (changes requested):
Looks good to me, only reservation is the `read_register/write_register` method names on the JTAG driver.

Would like to see them named something else to make it clearer that these are not the same as the common reg API methods implemented by higher-level drivers. Perhaps `shift_register` with a write/read switch argument might be better?

rust/pyapi/src/services/jtag.rs
coreyeng: This is most in-depth I've gotten with JTAG, and in much of the documentation I was looking at, the DR and IR were referred to as registers, even though they aren't registers in the usual sense. But, since I kept seeing them referred to as such, I added verify/write registers

Though, these will actually shift in/out the data and not leave the "register" in the same state. So probably a good update would be to either change the name as you suggest, or shift the assumed data in again so it maintains its state.

If we change to shift though, would we still want a shift_verify/write? Or just an option, as you had above? I'm fine with a name change, but would we want to keep the two-method approach consistent? I had planned to add a `capture` as well once I have that in order, but it could be a `capture_shift` or something instead?

ginty: I think the main thing here is to change the name to something else. From the point of view of a user who doesn't really understand protocols all that well, if they see methods called `write/verify_register` then they might rightly assume that they can give this a DUT register object and expect it to generate a valid transation.
So we should consider these as special method names which are essentially a DUT register handler trait and which the JTAG driver obviously doesn't support.

I don't have any special preference on what other name(s) to use, nor really about consistency between physical protocol drivers. If anything, I probably prefer physical drivers to implement their own unique APIs because it emphasises that at this level Origen is not providing much abstraction and the user really needs to know what they are doing.

coreyeng: Okeydoke. The functions to accept DUT register objects but its implied to be within the JTAG block. It was more for writing internal registers. I do already have a write/verify IR/DR so I guess that's sufficient then?

I don't actually use write/verify register anywhere. I added it for completeness, but going a bit too complete I guess. I just removed them unless you have another operation they should be replaced with. I think you can do everything you need to with just write/read IR/DR.

ginty: Yep I'm fine with that, I think those IR/DR read/write methods are all we need here.

ginty (approved):
Thanks for the update

PR #140, opened 2021-08-16
Prog gen

ginty:
This is a request to merge the work from the prog_gen branch into master so that it can be kept up to date with changes in master, e.g. update to lastest PY03.

As noted below I think it pretty much fully supprots the O1 flow control test case - https://github.com/Origen-SDK/origen_testers/blob/master/program/flow_control.rb

However, I seem to have misplaced the actual flow that I had converted to O2/python and so it currently has no test coverage :-(

This will be added in future whenever work resumes on the prog gen development.

Updates since the last prog gen drop:

* Now generates a referenced.list of patterns, same as O1
* Adds all the secondary files generated by O1 for V93k, e.g. .pmfl, .aiv, etc.
* Adds matching O1 v93k flow signature generation (unique ID appended to tests names)
* Adds almost full support for the O1 flow control APIs (I think), there is currently no testing of it but I believe this is very close to being able to generate O1s flow control API test cases
* Added unit_helpers.py:

```python
from origen.unit_helpers import *

flow.meas("read_pump", bin=119, hi_limit=45 * mV)
```


priyavadan (approved):
LGTM!

rust/origen/src/utility/file_actions.rs
```diff
@@ -10,6 +10,7 @@ enum Operation {
     InsertBefore,
     InsertBeforeAll,
     InsertAfter,
+    #[allow(dead_code)]
```
coreyeng, line 13: Is this needed? I don't see any other changes in this file that would need this.

coreyeng (approved):
Lots of useful stuff! Tried to look through as best I could and I don't see anything that looks problematic and since the light's green... LGTM!

rust/origen/src/utility/file_actions.rs
ginty: Yeah definitely needed to fix a build warning, I just checked master and the warning is there too, so that's why there is no other corresponding change here

PR #145, opened 2021-08-30
Metal Integration Waypoint #1

coreyeng:
Hello,

This is a first attempt at integrating `metal` as a dependency for `origen`. To start, I just did some very basic things to get my bearings and not bloat the PR.

### Origen Metal Dependency and More Version Verbosity

I added version info to [pyapi_metal](https://github.com/Origen-SDK/o2/compare/metal_integration?expand=1#diff-ea706d7f3f5ae71038f5b88e93bad44cfa6f8e29a6428d9971c865c49a08a905R6-R18) and [origen_metal](https://github.com/Origen-SDK/o2/compare/metal_integration?expand=1#diff-7d7679e60fa3d6807b907376dcdbb3027e7802a1d1cf1374bc171fe4f71cb295R14-R21). This is very minor, but as we have more moving parts, I think just having a record may be helpful while potentially allowing for version checks in a release script to ensure everything went alright. When run with `debug` verbosity, `origen -v` will spit out the `origen_metal` (Python), `pyapi_metal`, and `origen_metal` (Rust) versions:

image

I added `pyapi_metal` as a dependency to `pyapi`, allowing utilities to be used in both places. As this is meant only for non-pyclass/pymodule/etc. usage, this shouldn't cause library-size bloat nor duplicate definitions. All _python stuff_ should be used from `origen_metal._origen_metal` directly and not be reproduced in `_origen`.
I didn't see that `pyo3` supports `pathlib.Path` objects directly yet, so I just moved over the [pypath](https://github.com/Origen-SDK/o2/compare/metal_integration?expand=1#diff-ea706d7f3f5ae71038f5b88e93bad44cfa6f8e29a6428d9971c865c49a08a905R34-R47) macro I had as a very basic test case.

### Interactive Startup

I moved the interactive setup into `origen metal`, made it a bit more generic, and removed it from `origen`.  This is the same setup as before which just allows _console history_ to function in Windows.

### Python 3.9 Support

When I moved the interactive stuff, I also updated to [pyreadline3](https://pypi.org/project/pyreadline3/), which is a forked, active version of [pyreadline](https://github.com/pyreadline/pyreadline) that was blocking 3.9 support. Fixing this, 3.9 just worked. I updated the regression tests and release scripts to include 3.9. Haven't run the release scripts and am not 100% on [this](https://github.com/Origen-SDK/o2/compare/metal_integration?expand=1#diff-1630160ab3c5603f33112bc4c26e5ba56ce14653e4d795c38c195049c50d1b51R31), but based on the [python-setup action](https://github.com/actions/setup-python), this is my best guess.

### Python 3.10 Lookahead

Python 3.10 works alright in my environment with some prerelease Poetry version, but I could not for the life of me get the regression script to pass. In my environment, all is fine, but I can't reproduce the error. Python 3.10 is still unofficial, but [looks like its targeting October](https://www.python.org/dev/peps/pep-0619/#id5) to be official. I'll continue tinkering with this to see if we can get some lookahead. At the very least we'll need a Poetry update, but hopefully 1.2 will be out of prerelease by then. However, in debugging this, I did make two changes:

* Changed the [POETRY_VERSION](https://github.com/Origen-SDK/o2/compare/metal_integration?expand=1#diff-f33565d7e00780e326a7421954746ef901d297e2d887236f72377c19232adb71R17) to a `MINIMUM_POETRY_VERSION`.
* Prerelease versions seem to automatically resolve to _less than_ released versions, so when comparing versions, [I chuck out any prelease components](https://github.com/Origen-SDK/o2/compare/metal_integration?expand=1#diff-f33565d7e00780e326a7421954746ef901d297e2d887236f72377c19232adb71R240-R249). I only did this for the Poetry install, but we could move this upstream if needed.

### PyO3 0.14.4

[PyO3 0.14.4](https://pyo3.rs/v0.14.4/changelog.html) was released. Seems pretty minor, but I bumped up the version nonetheless.

### Minor Test Updates

A minor problem did come up with the regressions in the example app. When `origen_metal` was added as a local dependency, the `tests` directory there, which has an `__init__` and is therefore a module, superceeds the one in the test app. So, an `__init__` was added there to make it a module, but this caused some of the tests to fail as the _module path_ for these is now `tests` instead of empty. These were minor and expected given the needed update, but [just pointing them out](https://github.com/Origen-SDK/o2/compare/metal_integration?expand=1#diff-e66f19f22011ffedce224331d0e9769a0edc510187790c2b2fd2754357b6fc59R169-R213) regardless.

### SSL Requirement

In origen, I removed the SSL requirement from Windows builds since it requires a separate install for Windows and wasn't being used. I did the same for [metal](https://github.com/Origen-SDK/o2/compare/metal_integration?expand=1#diff-6f2aed686c1385919d86bf187b1130bc2d70df3acf13d961a983503c1e597589R20) but I can look at getting this to work again. I guess this built fine on the Windows machines during the regressions, I should probably give this some more thought in the future.

### Side Note

I plan to keep this branch around to work on generalizing and cleaning up things that could go in `metal`. I plan to do more PRs off of this branch as this occurs. If anything is known to be needed sooner, let me know and I can likely prioritize those items.

Thanks!


rust/pyapi_metal/Cargo.toml
```diff
@@ -8,20 +8,25 @@ authors = ["Origen-SDK"]
...
 
 [dependencies]
 origen_metal = { path = "../origen_metal" }
 
 # Statically links these libs to make the Python extension more portable
 libz-sys = { version = "1.0.25", features = ["static"] }
 curl-sys = { version = "0.4.34", features = ["static-curl"] }
+
+[target.x86_64-unknown-linux-musl.dependencies]
 openssl = { version = "0.10", features = ["vendored"] }
```
ginty, line 21: I'm not sure if this is correct, I don't think the linux-musl target is used by the origen_metal build flow currently, I think it should be `target.x86_64-unknown-linux-gnu.dependencies', or maybe we need another entry to add it for both?

ginty (changes requested):

rust/pyapi_metal/Cargo.toml
coreyeng: Good catch! This is the same as in `origen` too. I very likely copied this from SO back when we started in 2019 and wasn't 100% sure of the syntax and didn't even notice. Obviously we're not really using it then and it would've successfully blocked building in windows, which was all I was looking for!

Also looks like it was taken out of pyapi completely. I put it back in there with the same condition.

ginty: I'm not sure, but `origen` might need both, as the `musl` target is used (or at least used to be) for building the self-contained static CLI binary.
I think that doesn't really apply to an extension since it cannot be both self-contained and a library, so the unknown linux is used in combination with the manylinux checkers to ensure that any external linking is kept to a minimum.
Anyway, I'll approve this now since it is definitely fine now for origen_metal, thanks for making this update and for your support of origen_metal.

ginty (approved):

PR #146, opened 2021-09-01
Reference File API

ginty:
This moves Origen's functions for implementing reference file (change) tracking into metal, and adds a Python API for it.

Also removed Py3.6 from the build and regresssion.


coreyeng:
Looks good! Seems like relocating it origen metal wasn't too big a deal?

coreyeng (approved):

ginty:
Yeah that one was easy enough, and hopefully might be typical of what we will see - just needed to remove references to `origen.app.root` and replace with an API to allow the path(s) to be defined externally.

info-rchitect:
@ginty Was the concept of origen metal always something you thought origen would evolve to?  Just curious on how this all developed.

ginty:
> @ginty Was the concept of origen metal always something you thought origen would evolve to? Just curious on how this all developed.

@info-rchitect, nope, my life would be considerably easier right now if I had!

As you know I'm developing something Origen-like in AMD, but it's far too different to allow us to be able to use Origen directly - e.g. we're using our own backend model and vector generator.
Origen was very much designed with an all-or-nothing mindset and until now I've wanted to be able to use bits of Origen but have been un-able to because so much of it is coupled to Origen's application framework and the general assumption that if you are using bits of it then you are using all of it.

So the plan is to start moving chunks of Origen over to Origen Metal over time. If something's in Origen Metal it gives you the guarantee that it is generic and not coupled to Origen's application framework.

Ultimately I hope to see OM being the bigger of the two, with large components of Origen like the Program generator APIs and the model eventually moving over.

ginty:
Hi @coreyeng, I squeezed in an update to the differ to make it support multiple ignore blocks and fixed a bug I found with the original implementation.
I'll go ahead and merge though as I doubt you'll object to that.
Thanks!

PR #151, opened 2022-04-21
Add AST Infrastructure and STIL to Metal

ginty:
This is a big diff but is just refactoring without any functional changes.

The end goal was to make the STIL parser available in Origen Metal and to make the AST infrastructure (e.g. the Node type) more generic and also available from Metal.

The following has changed:

- The AST `Node` type is now generic and no longer coupled to the single node Attr type that previously encompassed all pattern and flow nodes. There are now 3 types of nodes used in O2 - `Node<PAT>` (Origen pattern), `Node<PGM>` (Origen test flow) and `Node<STIL>`.
- All of the generic AST stuff that was previously shared by the pattern and prog gens is now in `origen_metal::ast`
- The STIL parser is now in `origen_metal::stil`
- I ran into problems with having both `origen_metal::Error` and `origen::Error` types and so this also removes the Error type and all associates error macros from Origen, everything is now an `origen_metal::Error` (although it is aliased to `origen::Error`). Use the `bail!` macro anytime you want to generate and return an error from anywhere.
- The Python side now returns RuntimeErrors instead of OSErrors. I'm not honestly sure why that changed but RuntimeError seems like a better generic error to use and so I updated all the tests accordingly.


priyavadan:
quite a bit of updates... overall I think I follow what you have done here. Great work!

so keyring fails on Ubuntu-latest vs passing on Windows?

priyavadan (approved):
quite a bit of updates... overall I think I follow what you have done here. Great work!,  thank you... this will help us internally A LOT!

seems like regressions are failing on Ubuntu but passing on Windows ; related to  keyring?

ginty:
Yep something to so with this keyring crate, @coreyeng you got any thoughts/experience with this error?

It builds fine on my local machine, just not here.

priyavadan:
Maybe this will help 

from their page  (https://crates.io/crates/keyring/0.10.4) 

image

ginty:
Actually I think its related to the Rust version, I'm using a newer version that here and I can see it fail locally using the older version used in the regression test

coreyeng:
I had some issues with this as well and had the specs passing with [this toml](https://github.com/Origen-SDK/o2/blob/b91db1bfcdb56eded86c01d7e1175d08169c6c60/rust/origen/Cargo.toml#L82) and [this actions update](https://github.com/Origen-SDK/o2/blob/b91db1bfcdb56eded86c01d7e1175d08169c6c60/.github/workflows/regression_test.yml#L29-L36), which is along what Priyavadan had linked.

I did update Rust to 2021 but I've been only running Windows so far.

coreyeng:
Keyring is so far only used by the `users` though for password storage and is something I'll be working on if it fails for me when I get around to a PR. If you can't get it, I'm fine with you just removing it for now and I'll deal with it in my branch.

ginty:
Thanks @coreyeng
I've removed that patch and now running with latest rust and latest keyring seems to work fine,

coreyeng (approved):

PR #193, opened 2023-08-05
Rust update

coreyeng:
* Updated to rust 1.71.0 and updated dependencies that were giving warnings stating they contained other deprecated functionality. This is incompatible with older Rust versions though. Will need to update local Rust versions.
* Borrowed from @rlaj's branch to support stable rust. 

One note here is that one of the updated libraries didn't port over smoothly. `octocrab`, which has an API for Github, changed its API around. However, it now looks like it has some `actions` API, but this is non-spec-ed and will be revisited for the release procedure. [I've "todo-ed" it for now, with plans to revisit.](https://github.com/Origen-SDK/o2/compare/rust_update?expand=1#diff-d8b2d478db8da84023db368e797fb1d285c28ab7cac8361f83254a0cff4fe1bf)

This will cause some compile warnings which will be moot when the function is reimplemented.

### DIFF
--- a/python/origen_metal/tests/utils/test_mailer.py
+++ b/python/origen_metal/tests/utils/test_mailer.py
@@ -414,8 +414,7 @@ class TestMaillist(Common):
         ]
 
     def test_invalid_recipient_resolution(self):
-        invalid_email_err_msg = "Missing domain or user"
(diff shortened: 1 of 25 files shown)


rlaj (approved):
LGTM, thanks!

PR #201, opened 2024-06-13
STIL parser to keep all comments, not just annotations

rlaj:


ginty:
Good update, would like to do more with STIL post-processing in the future to catch when StimGen implements transactions out-of-order and keeping the SG comments around for that will be necessary

ginty (approved):
LGTM

PR #194, opened 2023-08-11
Feature/stil parser updates

priyavadan:
going to try and create this PR, moving to a newer rust should help with the regression failures

PR #202, opened 2024-06-14
STIL Parser Updates and OM AST Processor Refactor

rlaj:
This PR is an re-attempt of PR #194.  The merge conflicts have been cleaned up and a few more updates have been made.

Two main items: 

1. STIL parser updates
	-new method to pass options to the parser
	-various (minor) enhancements to handle different STIL versions and downstream requirements

2. OM AST Processor Refactor
	-removed recursive bits of the OM processor that was leading to stack overflow condition on some STIL use cases
	-refactor required some updates to the processors in origen core, specifically in the pattern generator


ginty (approved):
I'm fine with this, but should wait for @coreyeng to bless.

@coreyeng, a fmt seems to have been done at some point which makes this look bigger than it is.

Aside from what @rlaj notes, this also contains an update to `ast/node.rs` to eliminate recursion from the parser and make it loop instead.
This was necessary due to hitting a stack overflow when parsing some of our STILs.

ginty:
Thanks again for taking the pain to get us back onto the main line @rlaj!

coreyeng (approved):
Yep, looks good. Thanks! I've been trying to make a point to test everything in some capacity so if the tests work, I can't complain. We aren't yet, but I think we'll be using STIL more here in the future, so nice to have that back on the master!

Aside: I've not been good about keeping up with the formatter. I would also like to talk about the Python rules at some point (not urgently), as I'm not fan, particularly of the spacing rules. I find the continued line stuff really ugly, especially for wide-screen setups.

PR #203, opened 2024-06-26
STIL - Allow waveform defs with no semicolon

ginty:
AMD has historical STIL files where the final event in a waveform definition omits its `;` terminator.

This keeps standards-compliant STIL as the default and makes that legacy syntax an explicit runtime opt-in. Existing APIs and behavior remain unchanged:

```rust
let ast = origen_metal::stil::from_file(path)?;
```

AMD or another consumer can enable only the legacy final-event syntax per parser:

```rust
let ast = origen_metal::stil::Parser::new()
    .allow_missing_final_waveform_event_semicolon()
    .from_file(path)?;
```

The setting is propagated to included STIL files. Tests cover strict rejection, tagged and untagged waveform definitions, single- and multi-event definitions, include propagation, and equivalent AST output between standard and legacy forms.

### DIFF
--- a/rust/origen_metal/src/stil/mod.rs
+++ b/rust/origen_metal/src/stil/mod.rs
@@ -9,17 +9,101 @@ pub use nodes::STIL;
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 
-pub fn from_file(path: &Path) -> OrigenResult<Node<STIL>> {
-    let ast = parser::parse_file(path)?;
-    let load_path = vec![Path::new(path).parent().unwrap().to_path_buf()];
+#[derive(Clone, Copy, Debug, Default)]
+pub(crate) struct ParseOptions {
+    pub(crate) allow_missing_final_waveform_event_semicolon: bool,
+}
 
-    let ast = processors::includer::Includer::run(&ast, load_path, HashMap::new())?;
-    Ok(ast)
+/// A configurable STIL parser.
+///
+/// The default parser accepts only standard STIL syntax. Compatibility extensions must be
+/// explicitly enabled before parsing.
+#[derive(Clone, Debug, Default)]
+pub struct Parser {
+    options: ParseOptions,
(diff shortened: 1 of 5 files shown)


priyavadan (approved):
we discussed this , I have my thoughts but I understand the need/sentiment and am okay with the changes.

coreyeng:
Since it is outside the spec, would it be prudent to add an option for this (I don't mind which is the default)? Something like either `use_strict_stil_spec = true/false`, or `relax_stil_spec = true/false` and setup the parser accordingly?

coreyeng (changes requested):
Since it is outside the spec, would it be prudent to add an option for this (I don't mind which is the default)? Something like either `use_strict_stil_spec = true/false`, or `relax_stil_spec = true/false` and setup the parser accordingly?

rlaj (approved):
Approving, but have no problem with @coreyeng proposal.  Although, I think there are some other gaps the parser has w.r.t. to the spec, so `use_strict_stil_spec=true` might give a false sense of security.

ginty:
> Since it is outside the spec, would it be prudent to add an option for this (I don't mind which is the default)? Something like either `use_strict_stil_spec = true/false`, or `relax_stil_spec = true/false` and setup the parser accordingly?

Yeah that's fair, let me see if I can come up with a way to do this

coreyeng:
@rlaj, I'm not that familiar with STIL and you're probably right. I put that more in "we don't know what we don't know" category. If we do have an option though, we can lump other things we aren't thinking about in there. A "strict STIL parser" may be a WiP but we'll have that option to cover things we use that are outside of that.

ginty:
@rlaj, @coreyeng, this long-standing PR has finally been reworked to add an API to opt-in to the non-std STIL case found in some legacy AMD files.

coreyeng (approved):
Thanks! Looks good!

PR #207, opened 2024-08-14
Add GZ reader to STIL parser

rlaj:
Enable parsing of GZipped STIL files (*.gz)


rlaj:
Honestly hadn't considered Windows side :)  I'll take a look.

rlaj:
@coreyeng, regression seems OK.  Went ahead and added a gz-specific test for warm and fuzzies.

coreyeng (approved):
Thanks! Looks good to me.

PR #212, opened 2025-01-14
STIL: Make integer optional for scanin/scanout def

rlaj:
Fix ScanIn/ScanOut definition to make integer specification optional to match STIL spec


coreyeng:
PR looks good!

On the regressions, I think the ubuntu-latest changed up a lot of stuff. Everything was fine a week ago and I think the version changed. There's something [here](https://github.com/onnx/tensorflow-onnx/pull/2376/files#diff-4b9a66bfd30d4df0a48bdc4ef7d375a1b5f5ad4e98fd90188f674afe48694fe8) I was going to try to just see if it worked and at least get us back to passing. Nothing checked in recently should have broken Python 3.9+ and 3.7 on Linux.

coreyeng (approved):

PR #218, opened 2025-04-03
Add way to reset the test prog generator from Origen Metal

ginty:
Adds a way to reset the program generator and clear out all flows from Origen Metal.

~~I also re-instated the `[project]` section in the Origen metal `pyproject.toml` - we're no longer using Poetry and this is required to provide a standard interface to integrate into other envs for development.~~ Have left commented out based on the discussion below

Also added the following program generator updates:

* Adds much of the functionality from the O1 test_ids plugin. The algorithm and callback features that it supported have not been replicated here:
  ```python
    from origen_metal.prog_gen.test_ids import TestIDs
    
    # Create a new instance of TestIDs
    tids = TestIDs()
    
    # Restore a previously saved TestIDs instance
    tids = TestIDs("/path/to/my_test_ids.json")
    
    tids.include("bin", (100, 200))                  # Make a range of bin numbers available
    tids.include("number", (10000, 20000))           # Make a range of test numbers available
    
    tids.allocate("test1")    # => {"bin": 100, "number": 10000}
    tids.allocate("test1")    # => {"bin": 100, "number": 10000}   returns the same values
    tids.allocate("test2")    # => {"bin": 101, "number": 10001}
    # If manually assigning a bin number, let the allocator know so that it can remove it from the
    # pool of available bins
    tids.allocate("test3", bin=105)                  # => {"bin": 105, "number": 10002}
    
    # Assign a test number only
    tids.allocate("test4", bin=None, softbin=None)   # => {"number": 10003}
  ```
  I added the Python binding stuff in `origen_metal` itself which is cleaner/less duplication than adding an intermediate interface in `pyapi_metal` (we probably went down a wrong path there)
* Exposes the ability to configure the V93K SM7 test name uniqueness option: `origen_metal.prog_gen.set_uniqueness_option("none")`, also supports `"signature"` (the default) or `"flowname"`
* Makes a test invocation instance's name available to Python


coreyeng:
The `ProgGen` looks fine to me.

The `pyproject` update is concerning me a bit as I had to remove it for a similar reason - when I'd use pip directly, it'd get confused between that section and the poetry section. But, when I'd use poetry, it'd complain about the non-poetry section having undefined values that when I then added them, would say they were invalid. So, couldn't win.

What are you guys using there instead of Poetry? Poetry is decent, but IMO, it does seem to be a step backwards from Bundler. I actually wouldn't mind decoupling OM and Origen dev environments from Poetry. I think just a raw `pyenv` and `pip build` script would be sufficient for those and be clearer as to what's going on. More would need to be updated though to get the full tests and publishing working though, even for just OM.

I also just realized the LDAP tests were failing. I looked into that a bit and found my jumpcloud account was removed because I was just using my throwaway gmail account and not a 'real' work account. I'm looking back at forumsys and it seems to be up again and relatively good. I'll work on fixing those up, or at the very least, just make the failures warnings. I like the idea of having the LDAP regression tested if there's a good server to test against.

coreyeng (approved):

ginty:
> Watching the regressions, this is why I got away from it in the first place: image
> 
> That'll persists for a while after it first happens. Then it'll come back and be okay. Looks like a few other versions got through though.
> 
> If you want to merge this and we'll play "comment tag" in the pyproject section for a bit, that's fine with me. I don't see anything else that should affect this and I don't want to hold you up because my ldap tests are being difficult. Caveat being if you try to run the publish script Ronnie and I were using (the non-Maturin one), it'll rely on Poetry and may get lost.

Yeah, that's fine, thanks Corey, I'll back out your changes from this before merging, just wanted to see if it could pass.

ginty:
Hi @coreyeng, as ever with these things I ended up added more, could you please give this a final review?

coreyeng:
Yep! I've certainly been there before. I don't see anything that sticks out to me though. Looks good!

I'm wasn't able to put any time into the LDAP tests but I'll try to make some time soon for that to get our regressions passing again. I still don't want to hold you up though if you want to merge this into master, I'll pull it down and build in the new hooks for the LDAP testing.

PR #221, opened 2026-01-19
Windows cli usability

pderouen:
Some very minor usability updates to the CLI:

- In Windows dev mode the "\\" in the path string inserted in the TOML is interpreted as an escape character
- The paths in strings passed to Python for interpretation that include double quote characters are also interpreted as escape characters

In both cases changing any "\\" to "/" works for both operating systems. Not sure why Rust <code>PathBuf</code> doesn't have a built-in way to get linux style path separators.

I'm not sure what the deal is with the regression checks. The CLI checks near the end of the flow are taking several hours to complete. So, the job is getting cancelled.

### DIFF
--- a/test_apps/no_workspace/templates/pyproject.toml
+++ b/test_apps/no_workspace/templates/pyproject.toml
@@ -7,6 +7,19 @@ authors = ["Origen-SDK"]
 [tool.poetry.dependencies]
 python = ">=3.7.0,<3.13"
 docutils = "!=0.21.post1"
+
+# Pin dependencies that have dropped Python 3.7 support (saves 30 minutes of dependency resolving)
+# These versions are the last to support 3.7 while still working on 3.12
+more-itertools = ">=8.0.0,<10.0.0"
+zipp = ">=3.4.0,<3.16.0"
+setuptools = ">=40.0.0,<70.0.0"
+importlib-metadata = ">=4.0.0,<7.0.0"
+filelock = ">=3.0.0,<3.13.0"
+pygments = ">=2.7.0,<2.18.0"
+virtualenv = ">=20.0.0,<20.27.0"
+platformdirs = ">=2.0.0,<4.0.0"
+distlib = ">=0.3.0,<1.0.0"
+
 {% if local_origen %}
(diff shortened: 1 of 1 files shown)


pderouen:
It's doing the same thing again. The "run global" checks are taking forever to run.

pderouen:
> Would appreciate if we can maintain Py 3.7 support. Understood that its becoming a pain and everyone else is moving on, but we're on semiconductor environment timelines. We should be able to drop it next year some time though.

@ginty, understood. It's not a 3.7 issue. It's the poetry dependency resolver taking ~30 minutes for each global test when running in 3.7, leading to hitting the job run time limit. I'm trying a few things to speed it up.

info-rchitect:
> > Would appreciate if we can maintain Py 3.7 support. Understood that its becoming a pain and everyone else is moving on, but we're on semiconductor environment timelines. We should be able to drop it next year some time though.
> 
> @ginty, understood. It's not a 3.7 issue. It's the poetry dependency resolver taking ~30 minutes for each global test when running in 3.7, leading to hitting the job run time limit. I'm trying a few things to speed it up.

Why not just convert to uv given the long poetry runtimes?

pderouen:
@info-rchitect, seems like a very good option. It's outside the scope of this PR though... I just want to be able to use the tab key to autocomplete file names and paths from windows without getting run time errors. I don't mind letting this PR wait until the 3.7 global test run duration issue is resolved.

rust/origen/cli/src/commands/env.rs
```diff
@@ -311,3 +316,11 @@ fn delete_virtual_env() {
         }
     }
 }
+
+fn normalize_path_string(path: &PathBuf) -> String {
+    if cfg!(windows) {
+        path.display().to_string().trim_start_matches(r"\\?\").to_string().replace("\\", "/")
```
coreyeng: This is the same sort of thing where technically you might misinterpret a valid path. But I'm not sure how well Windows even supports that. It might just be a note for future (if it ever comes up) that whacky windows paths aren't supported.

That said, even powershell vs. CMD can treat path differently so... not sure how much we should care.

PR #224, opened 2026-02-24
Add UFLEX tester

sekharsankar:
Tester updates- created base class for modularity between j750 and uflex


ginty:
Hi @sekharsankar, thanks for the PR.

Could you merge the latest master into this branch please, that should now trigger the regression

pderouen:
I clicked the allow to run button. But, there is some sort of setup issue: https://github.com/Origen-SDK/o2/actions/runs/23605265816

The run from the forked repo: https://github.com/sekharsankar/o2/actions/runs/23568809998

pderouen:
@ginty it seems our repository variables are not readable when the GA checks are run on this PR (maybe because it's a fork). It's complaining about the supported os variable. We also ran into this when trying to get the checks to run on the forked repo. I had to get the variable definitions to him. Those don't come across with a fork.

@sekharsankar what were the variables we had to define in your forked repo?

sekharsankar:
@pderouen, we had to define PYTHON_VERSIONS, RUST_VERSION and SUPPORTED_OS in my forked repo

ginty:
We should definitely fix that, but shouldn't couple to this PR.

I can see its passing in the fork and none of this is an area that AMD use, so I'm fine with it, LGTM.

coreyeng (approved):
I'm sure GitHub has some means to support it (can see if the AI chat thing has a recommendation), or we can move the variable declarations to new action, run that action, and use them in the others.

I moved those pieces into variables so I could use them across the release and regression actions - only making Python releases for versions/OS were testing, and making sure the built libraries were using the same Rust version as the regressions.

Definitely that can be cleaned up but I (or Paul or Shekhar, if you have time) can make another fork to replicate so I agree, we don't need to couple these. The regressions ultimately ran and passed so I'm happy.

This is related, but also not necessary to block this, but should we distinguish between a UFLEX and UFLEX+? We've got a few small differences (mostly pattern compiler related) between them, and I've got adding a uflex_plus target to O1 on my 'eventually' list.

pderouen:
> This is related, but also not necessary to block this, but should we distinguish between a UFLEX and UFLEX+? We've got a few small differences (mostly pattern compiler related) between them, and I've got adding a uflex_plus target to O1 on my 'eventually' list.

The way I handled UF+ in O1 was to allow the [digital instrument](https://origen-sdk.org/origen/guides/pattern/ultraflex/#Digital_Instrument) to be specified. That was all I needed to get direct compile working. Can do something similar here. I was thinking to address this in a future update. We will also want digsrc/digcap supported similar to O1 eventually.

sekharsankar:
Right now I have defaulted the dig instrument header to be "hsdm". For UF+, we would just need that to be "hsdp"

pderouen:
> Right now I have defaulted the dig instrument header to be "hsdm". For UF+, we would just need that to be "hsdp"

Yes, but after talking with Corey, I realized he's talking about the compiling into a binary file (their group has Origen do this for them).

pderouen (approved):
Looks good to me as well.

PR #227, opened 2026-07-06
Fix SMT8 pass/fail flow rendering bug

ginty:
# Fix SMT8 Pass/Fail Flow Rendering Bug

## Summary

This PR fixes an SMT8 flow rendering bug where a conditional retest could be guarded by its own pass/fail result before it had executed.

The failure occurred when a test was conditionally executed based on a prior test result, and that conditional test also had its own result branch. In that shape, the SMT8 renderer reused a single execute-line buffer for nested result branches. The nested branch could flush the parent branch contents under the nested test's `.pass` guard, changing the generated flow semantics.

Before:

```java
primary_check.execute();
if (!primary_check_retest.pass) {
    primary_check_retest.execute();
}
```

After:

```java
primary_check.execute();
if (!primary_check.pass) {
    primary_check_retest.execute();
    if (!primary_check_retest.pass) {
        ...
    }
}
```

## Changes

- Replaced the single SMT8 execute-line buffer with a stack of execute-line buffers.
- Result branch rendering now isolates nested `OnPassed` / `OnFailed` bodies and flushes captured lines back to the active parent buffer when needed.
- Refactored repeated SMT8 result-branch rendering code into a local helper so the `Group`, `Test`, and `TestStr` paths share the same buffering and guard-emission logic.
- Preserves and restores the previous `render_bins` state while rendering nested test result branches.
- Pinned Windows CI jobs from `windows-latest` to `windows-2022` and generalized Windows workflow conditionals to avoid runner-image drift that was breaking the PR before the SMT8 tests could run.
- Added a focused SMT8 regression test using `TestStr` nodes to avoid external test-template dependencies while still exercising the full program-generation processing pipeline.

## Verification

Ran:

```bash
cargo test flow_generator --lib
cargo test nested_result_branch_keeps_parent_guard --lib
```

Also verified the updated workflow files parse as YAML:

```bash
ruby -e "require 'yaml'; %w[.github/workflows/regression_test.yml .github/workflows/publish.yml .github/workflows/publish_metal.yml].each { |f| YAML.load_file(f, aliases: true); puts \"OK #{f}\" }"
```

## Notes

- This does not change the public API.
- The functional fix remains local to SMT8 flow rendering.
- The Windows CI change is an environment stabilization change for GitHub Actions, not a runtime/product behavior change.
- The existing flag optimization remains intact; the issue was the renderer's shared buffer state during nested result-branch emission.

### DIFF
--- a/rust/origen_metal/src/prog_gen/advantest/smt8/processors/flow_generator.rs
+++ b/rust/origen_metal/src/prog_gen/advantest/smt8/processors/flow_generator.rs
@@ -203,6 +203,55 @@ impl FlowGenerator {
         }
     }
 
+    fn render_result_branch(
+        &mut self,
+        branch: &Node<PGM>,
+        subject_name: &str,
+        suppress_bins: bool,
+    ) -> Result<()> {
+        let guard = match &branch.attrs {
+            PGM::OnPassed(_) => Some(format!("if ({}.pass) {{", subject_name)),
+            PGM::OnFailed(_) => Some(format!("if (!{}.pass) {{", subject_name)),
+            _ => None,
+        };
+
+        if let Some(guard) = guard {
+            let original_render_bins = {
+                let current_flow = self.flow_stack.last_mut().unwrap();
+                let original_render_bins = current_flow.render_bins;
+                current_flow.start_execute_buffer();
+                if suppress_bins {
+                    current_flow.render_bins = false;
+                }
+                original_render_bins
+            };
+            branch.process_children(self)?;
+            {
+                let current_flow = self.flow_stack.last_mut().unwrap();
+                current_flow.render_bins = original_render_bins;
+                let buffered_lines = current_flow.finish_execute_buffer();
+                if !buffered_lines.is_empty() {
+                    current_flow.execute_line(guard);
+                    current_flow.flush_buffered_execute_lines(buffered_lines, 1);
+                    current_flow.execute_line("}".to_string());
+                }
+            }
+        }
+        Ok(())
+    }
+
+    fn render_result_branches(
+        &mut self,
+        children: &Vec<Box<Node<PGM>>>,
+        subject_name: &str,
+        suppress_bins: bool,
+    ) -> Result<()> {
+        for branch in children {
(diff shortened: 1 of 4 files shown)


rust/origen_metal/src/prog_gen/advantest/smt8/processors/flow_generator.rs
```diff
@@ -689,31 +700,41 @@ impl Processor<PGM> for FlowGenerator {
                     {
                         for n in &node.children {
                             if matches!(n.attrs, PGM::OnPassed(_)) {
-                                self.flow_stack.last_mut().unwrap().buffer_execute_lines = true;
-                                self.flow_stack.last_mut().unwrap().render_bins = false;
+                                let original_render_bins = {
```
priyavadan: should we move this to a helper method? its repeated multiple times  below

rust/origen_metal/src/prog_gen/advantest/smt8/processors/flow_generator.rs
```diff
@@ -689,31 +700,41 @@ impl Processor<PGM> for FlowGenerator {
...
+                                    current_flow.render_bins = false;
+                                    original_render_bins
+                                };
                                 n.process_children(self)?;
                                 {
                                     let current_flow = self.flow_stack.last_mut().unwrap(); 
-                                    current_flow.buffer_execute_lines = false;
-                                    current_flow.render_bins = true;
-                                    if !current_flow.execute_lines_buffer.is_empty() {
+                                    current_flow.render_bins = original_render_bins;
```
priyavadan: similar comment to earlier

rust/origen_metal/src/prog_gen/advantest/smt8/processors/flow_generator.rs
ginty: Done

rust/origen_metal/src/prog_gen/advantest/smt8/processors/flow_generator.rs
ginty: Done

PR #229, opened 2026-07-30
STIL Parser Bug Fixes and Vector Comment Regression Fix

rlaj:
### Background

A prior optimization commit merged the separate `vector_with_comment` variant into a single `vector = ${ ... ~ vector_comment? }` rule. The subsequent "quick patch" reverted `vector` back to a plain `{ }` rule to fix a regression where internal `//` and `/* */` comments inside Vector blocks caused parse failures (because the `$` modifier suppresses implicit COMMENT handling). That revert was correct to fix the internal-comment bug, but it introduced a new regression: trailing same-line comments (`} // comment`) were no longer captured as **children** of the Vector AST node — they fell through to the outer implicit COMMENT rule and became siblings instead.

---

### Changes

**`stil.pest`**

Restored `vector_comment?` capture with a properly structured compound-atomic rule that avoids both failure modes:

```pest
// Before (quick patch — lost vector_comment as AST child):
vector = { "V" ~ "ector"? ~ "{" ~ (cyclized_data | non_cyclized_data)* ~ "}" }

// After:
vector = ${ "V" ~ "ector"? ~ s ~ "{" ~ (s ~ (cyclized_data | non_cyclized_data))* ~ s ~ "}" ~ vector_comment? }
cyclized_data    = !{ (sigref_expr ~ "=" ~ vec_data+ ~ EOS) | ... }
non_cyclized_data = !{ (time_value ~ sigref_expr ~ "=" ~ event_list ~ EOS) | ... }
```

Key mechanics:
- `$` (compound-atomic) prevents implicit WHITESPACE between `}` and `vector_comment?`, ensuring only a same-line trailing comment is captured as a vector child — a comment on a subsequent line is correctly left for the outer context.
- The existing silent `s = _{ (space | N | one_line_comment | block_comment)* }` rule is used in whitespace positions inside `$`, correctly handling both newlines and comments (line and block) between `cyclized_data` entries.
- `cyclized_data` and `non_cyclized_data` are marked `!` (non-atomic) to re-enable implicit WHITESPACE for their own `~` sequences when called from the `$` vector context. Without this, pest's atomicity propagation from `$` to sub-rules suppresses the implicit WHITESPACE needed to match `sigref_expr = vec_data;` patterns with spaces.

**`parser.rs`**

- Fixed `Rule::call` to apply `unquote()` to the procedure name, matching the handling of `macro_statement`, `procedures_def`, and `macro_def`. Previously, a `Call "quoted_name" { ... }` would produce an AST node with the literal quoted string including surrounding `"` characters.

**`test_apps/python_app/vendor/stil/example17.stil`** (new)

New test fixture exercising:
- Vectors with a trailing same-line `// comment`
- Vectors with internal `//` line comments and `/* */` block comments between data entries
- Plain vectors without any trailing comment (baseline)

**`parser.rs` — new tests**

- `test_example17_can_parse` — parse coverage for the new fixture
- `test_vector_inline_comment_is_child_of_vector` — walks the AST and asserts that a trailing `// comment` on the same line as `}` is stored as a direct `STIL::Comment` child of the `STIL::Vector` node
- `test_vector_with_internal_comments_parses` — asserts that internal `//` and `/* */` comments between `cyclized_data` entries parse correctly AND the trailing comment is still a Vector child

All 27 STIL parser tests pass.

### DIFF
--- a/rust/origen_metal/src/stil/parser.rs
+++ b/rust/origen_metal/src/stil/parser.rs
@@ -1008,7 +1008,7 @@ pub fn to_ast(mut pair: Pair<Rule>, source_file: Option<&str>) -> Result<AST<STI
             Rule::call => {
                 let mut p = pair.into_inner();
                 ids.push(
-                    ast.push_and_open(node!(STIL::Call, p.next().unwrap().as_str().to_string())),
+                    ast.push_and_open(node!(STIL::Call, unquote(p.next().unwrap().as_str()))),
                 );
                 pairs.push(p);
             }
--- a/rust/origen_metal/src/stil/parser.rs
+++ b/rust/origen_metal/src/stil/parser.rs
@@ -1344,4 +1344,65 @@ mod tests {
             Err(e) => { println!("{}", e); assert_eq!(1, 0); }
         }
     }
+
+    #[test]
+    fn test_example17_can_parse() {
+        let txt = read("example17");
+        match STILParser::parse(Rule::stil_source, &txt) {
+            Ok(_) => {}
+            Err(e) => { println!("{}", e); assert_eq!(1, 0); }
+        }
+    }
+
+    /// Verifies that a trailing same-line `// comment` after a Vector block's closing `}`
+    /// is captured as a direct child Comment node of the Vector AST node, not as a sibling.
+    #[test]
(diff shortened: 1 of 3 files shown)


pderouen (approved):
lgtm

PR #231, opened 2026-07-31
Expand PGM AST with generic flow nodes, conditions, and Model extensions

rlaj:
## Summary

Adds a set of generic, additive nodes and conditions to the PGM AST to better support arbitrary test program import, export, and round-trip scenarios. Also adds a `custom_data` metadata field and a `Default` impl to `Model` for use in import pipelines not tied to a specific tester.

## Why

The PGM AST was missing several fundamental constructs that appear across all major tester formats. Without them, importers had to misuse `PGM::Render` as a passthrough vehicle, losing semantic information and making round-trip fidelity impossible. The gaps fell into three categories:

- **Variable and parameter nodes** — the existing `Variable` in the Model tracked flow-control flags; there was no node for typed data variable declarations or flow parameters as they appear inline in a program's source representation.
- **Flow-control nodes** — no structured representation for bounded loops, labeled jumps, external procedure calls, callbacks, or the error/abort result handler (complementing the existing `OnPassed`/`OnFailed`).
- **Condition and annotation nodes** — no site-number conditions, no bin-based conditions, no expression-based conditions, no structured comments, and no way to store format-level metadata on the Model.

The `custom_data` field on `Model` and its `Default` impl address a secondary gap: import pipelines that assemble a model outside the normal tester-targeted flow had no convenient construction path and no generic storage for format-level metadata.

## Breaking Changes & Linked Issues

**Breaking changes:** None. All changes are additive enum variants on `PGM` and `FlowCondition`. Every existing processor and validator in `prog_gen/processors/`, `prog_gen/validators/`, `prog_gen/advantest/`, and `prog_gen/teradyne/` already uses `_ =>` catch-all arms; no exhaustive match required updating. The `custom_data` field is a new public field initialized in `Model::new()` with no effect on callers. The `Default` impl is purely additive.

**Linked issues:** None.

## Notes for Reviewers

The `GroupType` enum (`Flow`, `Test`) was intentionally left unchanged despite being in scope. With only two variants it is very likely matched exhaustively in processor code; adding variants without auditing every match site would be a breaking change. That extension is tracked separately.

The `Loop` node uses `Option<u32>` for iteration count (`None` = infinite/condition-driven) rather than a `FlowCondition` payload. This keeps `Loop` independent of the condition system — a while-loop condition can be expressed as a `Condition` node wrapping the loop body. Reviewers should confirm this decomposition is consistent with how they'd want generators to consume it.

`IfBin`/`UnlessBin` reuse the existing `BinType` enum (`Good`/`Bad`). If finer-grained bin matching (e.g. by soft bin number range) is needed in future, this variant would need to be extended or replaced — the current form covers the common case of "did this DUT pass or fail its bin assignment."

### DIFF
--- a/rust/origen_metal/src/prog_gen/model/mod.rs
+++ b/rust/origen_metal/src/prog_gen/model/mod.rs
@@ -81,7 +81,7 @@ pub enum FlowCondition {
     IfAllSitesFlag(Vec<String>),
     /// A raw tester-specific expression that doesn't map to any of the structured condition
     /// variants above. Processors that don't understand the expression should pass it through
-    /// unchanged. Used for conditions like TestStand expression strings.
+    /// unchanged. Used for raw tester-specific expression strings.
     IfExpr(String),
     /// The negated form of `IfExpr`. Passes when the expression evaluates to false.
     /// Processors that don't understand the expression should pass it through unchanged.
--- a/rust/origen_metal/src/prog_gen/model/mod.rs
+++ b/rust/origen_metal/src/prog_gen/model/mod.rs
@@ -79,6 +79,41 @@ pub enum FlowCondition {
     UnlessFlag(Vec<String>),
     IfAnySitesFlag(Vec<String>),
     IfAllSitesFlag(Vec<String>),
+    /// A raw tester-specific expression that doesn't map to any of the structured condition
+    /// variants above. Processors that don't understand the expression should pass it through
+    /// unchanged. Used for conditions like TestStand expression strings.
+    IfExpr(String),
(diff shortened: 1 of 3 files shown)


pderouen (approved):
I appreciate the relatively small amount of changes to review. Will there be any tests for these new nodes in future PRs?

rlaj:
> I appreciate the relatively small amount of changes to review. Will there be any tests for these new nodes in future PRs?

Possibly.  Didn't think to add them here since this is only new data types.  The tests will come when they start being used in processors.

